PR #9, opened 2017-03-07
Add support for LLVM3.5 era function attributes.

joehendrix:
This adds support for some function attributes (those in LLVM 3.5 to AST while minimizing changes to Text.LLVM interface.


elliottt:
Thanks Joe!

PR #12, opened 2017-04-20
Warning fix/version bump

robdockins:


elliottt:
Shouldn't the version be bumped to 0.8.0.0, or was this just for the additional export?

robdockins:
The version bump is for the changes to the data layout parsing and datastructures.  If you think it should be for a more significant version number, that's fine with me.

EDIT: To be clear, the concern I want to address is this: I want an incremented version number so I can mention it in the version bounds of downstream packages.  This increment doesn't necessarily need to coincide with a fresh public Hackage release of llvm-pretty; therefore, it might make sense not to bump anything above the lowest number until a public release is prepared (to avoid the possibility of wierd sequence gaps in the public version numbers).

elliottt:
Ah, OK. As long as we don't forget to bump it to `0.8.0.0` when we actually do the next release, I'm fine with `0.7.1.1`.

PR #19, opened 2017-11-06
Add lens names "fieldNameLens" using microlens

TomMD:
Microlens is lens-compatible and has a small fraction of the build cost, suitable for library dependencies assuming the TH dep is OK.


elliottt:
Looks good! Were you thinking that we should consider doing a release soon?

TomMD:
Soon but not yet, I think (unless someone else pushes for it).  I hope to extract some really basic function that are used both by my work and Crucible (ex: `declarationFromDefinition`).  I'm not sure if these belong in a `Text.LLVM.Rewrite` module or somewhere else but that's an issue for another day.

PR #22, opened 2018-03-01
Print identifiers with special characters as escaped quoted strings.

brianhuffman:
This behavior matches the LLVM Language Reference Manual:

"Named values are represented as a string of characters with their
prefix. For example, %foo, @DivisionByZero, %a.really.long.identifier.
The actual regular expression used is ‘[%@][-a-zA-Z$._][-a-zA-Z$._0-9]*‘.
Identifiers that require other characters in their names can be
surrounded with quotes. Special characters may be escaped using "\xx"
where xx is the ASCII code for the character in hexadecimal...
Unnamed values are represented as an unsigned numeric value with their
prefix. For example, %12, @2, %44."

### DIFF
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -19,7 +19,7 @@ module Text.LLVM.PP where
 import Text.LLVM.AST
 
 import Control.Applicative ((<|>))
-import Data.Char (isAscii,isPrint,ord,toUpper)
+import Data.Char (isAlphaNum,isAscii,isDigit,isPrint,ord,toUpper)
 import Data.List (intersperse)
 import qualified Data.Map as Map
 import Data.Maybe (catMaybes,fromMaybe,isJust)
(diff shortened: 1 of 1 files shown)


elliottt:
Thanks Brian! Are you still building from this repository, or do you need a hackage release soon?

brianhuffman:
We're still building directly from this repo, so we're not waiting on a release.

brianhuffman:
While creating this commit, I noticed that the string escaping function `ppStringLiteral` implicitly assumes that all characters are in the range 0x00 to 0xff. Perhaps `Symbol` and `Ident` should use `ByteString` instead of `String`? See GaloisInc/llvm-pretty-bc-parser#63

elliottt:
That seems fine to me. If you're interested in pushing that change through that would be great, otherwise I'll make an issue so it doesn't get lost.

brianhuffman:
I'll make another pull request for it. Or you could just give me push access if that's easier for you.

elliottt:
You've actually got a pending invitation for commit access :)

brianhuffman:
I have no idea where to find that invitation, and I've been looking for 10 minutes :(

EDIT: I finally found it in my personal email (although the accept invitation link went to a 404 page at first, until I pasted the link into a different browser where I was logged in to github)

elliottt:
Well done github!

PR #27, opened 2018-10-13
Generalize updateAliases to functions with applicative codomain

langston-barrett:
This may seem a little obtuse but I promise it arose for a good reason :smile: 

I use `coerce` to avoid some runtime overhead of `Identity` (un)wrapping.


elliottt:
Thanks!

PR #37, opened 2018-11-28
Add DWARF address space to DIDerivedType (added in LLVM 6)

langston-barrett:
We should make an issue to track that there is no support for printing this yet.

LLVM review: https://reviews.llvm.org/D29670
Github mirror commit: https://github.com/llvm-mirror/llvm/commit/2cee5cc825969057f8d23b507a4861e606216aed
It's an unsigned int: https://github.com/llvm-mirror/llvm/blob/release_60/include/llvm/IR/DebugInfoMetadata.h#L786


langston-barrett:
Closing to minimize the Fryingpan build matrix.

PR #50, opened 2018-12-04
Always print "distinct" for DICompileUnit and DISubprogram

### LINKED ISSUE #47: Mandatory "distinct" for DISubprogram and DICompileUnit
When attempting to re-assemble this file with `llvm-as` from LLVM 6 after disassembling it with `llvm-pretty`/`llvm-pretty-bc-parser`, I get the following complaint:
```
$ nix-shell --pure -p llvm_6 --run "llvm-as /run/user/1000/cfi-eof-prologue.newllvm-disasm24413-2.ll"                                                        
llvm-as: /run/user/1000/cfi-eof-prologue.newllvm-disasm24413-2.ll:158:40: error: missing 'distinct', required for !DICompileUnit                             
                       templateParams: !DICompileUnit(language: 4, file: !3,
                                       ^
```
Unfortunately, my first attempt to fix this (putting "distinct" in the indicated location) failed:
```
$ nix-shell --pure -p llvm_6 --run "llvm-as /run/user/1000/cfi-eof-prologue.newllvm-disasm24413-2.ll"                                                        
llvm-as: /run/user/1000/cfi-eof-prologue.newllvm-disasm24413-2.ll:158:40: error: expected metadata operand                                                   
                       templateParams: distinct !DICompileUnit(language: 4, file: !3,  
```
Notice that in [the relevant LLVM commit](https://github.com/llvm-mirror/llvm/commit/c61bc48acb6bd55fa6b96b9b27de079b441e0b17), they never have a `DICompileUnit` inline. My hypothesis is that this is a bug in their parser (they assume they never see `DICompileUnit` inline). However, we might just want to always output those on their own line.

langston-barrett:
This has been tested with llvm-pretty-bc-parser, and with this change, the output of the printer/parser roundtrip will be successfully re-assembled by LLVM 6 in situations where it used to output invalid assembly. In particular, the roundtrip succeeds for the following file:

https://github.com/llvm-mirror/llvm/blob/master/test/DebugInfo/AArch64/cfi-eof-prologue.ll

Fixes #47

### DIFF
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -44,6 +44,15 @@ data Config = Config { cfgLoadImplicitType :: Bool
                        -- instruction is implied.
 
                      , cfgUseDILocation :: Bool
+
+                     , cfgDistinctDICU :: Bool
+                     -- ^ After LLVM 3.7, @DICompileUnit@ must be "distinct".
+                     -- If this is set, the pretty-printer will only produce
+                     -- valid assembly if the AST doesn't have inline
+                     -- @DICompileUnit@ or @DISubprogram@, since can't handle
+                     -- inline @distinct@ declarations.
+                     --
+                     -- https://github.com/llvm-mirror/llvm/commit/c61bc48acb6bd55fa6b96b9b27de079b441e0b17
                      }
 
 withConfig :: Config -> (LLVM => a) -> a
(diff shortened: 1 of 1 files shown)


langston-barrett:
I don't think this is the right approach, I think instead that clients of this library should have well-formed ASTs

PR #52, opened 2018-12-05
Add compile unit field to DISubprogram

### LINKED ISSUE #51: Add field Unit to DISubprogram
In 2016, several parameters to `DICompileUnit` were shifted back by one index (Why do that, rather than add it on to the end, in the backwards-compatible way, like they do for so many other new fields? :thinking:).

That commit also [makes the `dicuSubprograms` field unused](https://github.com/llvm-mirror/llvm/commit/4eeaa0da042055b1cd17b339e2dde518a3026033#diff-f26e75cce0f345af226cb16afcd749cbR1101), but that's a non-breaking change.

##  `Unit`

 - Record index: `15`
 - Type: [`Maybe PValMd`](https://github.com/llvm-mirror/llvm/blame/release_40/lib/Bitcode/Reader/MetadataLoader.cpp#L1279)
 - Default: `Nothing`
 - Git commit: https://github.com/llvm-mirror/llvm/commit/4eeaa0da042055b1cd17b339e2dde518a3026033#diff-7f530e364c7ddf1ea5bfa80d03d88c6dR1388
 - LLVM review: https://reviews.llvm.org/D19034

langston-barrett:
Fixes #51

### DIFF
--- a/llvm-pretty.cabal
+++ b/llvm-pretty.cabal
@@ -1,5 +1,5 @@
 Name:                llvm-pretty
-Version:             0.7.4
+Version:             0.7.5
 License:             BSD3
 License-file:        LICENSE
 Author:              Trevor Elliott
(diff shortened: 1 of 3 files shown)


langston-barrett:
This has been tested against [the corresponding change in the parser](https://github.com/GaloisInc/llvm-pretty-bc-parser/compare/master...siddharthist:disp-unit?expand=1).

PR #54, opened 2018-12-07
Fix DITemplateTypeParameter and DITemplateValueParameter

langston-barrett:
Tested and working on a _very_ large C++ program.

### DIFF
--- a/llvm-pretty.cabal
+++ b/llvm-pretty.cabal
@@ -1,5 +1,5 @@
 Name:                llvm-pretty
-Version:             0.7.6
+Version:             0.7.7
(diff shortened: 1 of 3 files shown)


langston-barrett:
I ran `git tag 0.7.7` before I pushed and ran `git push --tags`, but I think those may only have effects in my fork. After this is merged, I might push a tag to this remote.

langston-barrett:
I created a tag: https://github.com/elliottt/llvm-pretty/releases/tag/llvm-pretty-0.7.7

I probably should've been doing that with each change...

PR #55, opened 2018-12-07
Bump version to 0.8.0

langston-barrett:
The Package Version Policy (https://pvp.haskell.org/) dictates that
any API-breaking change should result in a bump of the major version number.
Unfortunately, I violated this in several of my previous PRs, bumping
only the minor versions even when the changes were breaking. This is an
attempt to patch that mistake up.


langston-barrett:
Added a git tag: https://github.com/elliottt/llvm-pretty/releases/tag/llvm-pretty-0.8.0

PR #60, opened 2019-01-10
Fix typo in data constructor

robdockins:


langston-barrett:
This will (of course) require downstream changes in `llvm-pretty-bc-parser` and `crucible-llvm` (which is why I hadn't done it earlier). But I'm all for it!

langston-barrett:
It's probably worth bumping the Cabal minor version number, since this changes a constructor of an exported datatype.

PR #61, opened 2019-01-15
Return more information from an invalid IndexResult

langston-barrett:


langston-barrett:
Closing to minimize the Fryingpan build matrix.

PR #64, opened 2019-05-14
Use Template Haskell to generate HasLabel instances

brianhuffman:
Switching from GHC.Generics to Template Haskell significantly reduces the memory footprint and execution time of `relabel` operations. See also GaloisInc/llvm-pretty-bc-parser#129.


elliottt (approved):
Looks great! I'm surprised that the GHasLabel stuff didn't go away through optimization -- thanks for fixing this!

PR #65, opened 2019-07-23
Add support for X86_fp80 (long double) 80-bit float type.

kquick:


src/Text/LLVM/AST.hs
```diff
@@ -326,8 +340,10 @@ primTypeNull (FloatType ft) = floatTypeNull ft
 primTypeNull _              = ValZeroInit
 
 floatTypeNull :: FloatType -> Value' lab
-floatTypeNull Float = ValFloat 0
-floatTypeNull _     = ValDouble 0 -- XXX not sure about this
+floatTypeNull Float    = ValFloat 0
+floatTypeNull Double   = ValDouble 0 -- XXX not sure about this
+floatTypeNull X86_fp80 = ValFP80 $ FP80_LongDouble 0 0
+floatTypeNull _        = error "must be a float type"
```
langston-barrett, line 346: Could we instead make this exhaustive? (I'm sure you considered that, just curious why it's not possible)

langston-barrett (approved):

langston-barrett:
Thanks @kquick!

src/Text/LLVM/AST.hs
kquick: @siddharthist That would be possible, I was just focused on X86_fp80 support at this stage.  

It wasn't clear to me what the XXX comment was intending.  Presumably it was a reference to the generic use of ValDouble for all other floating types (Half, FP128, FP128_PPC, FP80) is not appropriate.  If that presumption is correct, then absent an actual implementation appropriate for those, the error comment is probably an improvement, and I'll create an issue to specifically address the other types in the future.

PR #71, opened 2019-10-17
Fix build with GHC 8.8.1

atomb:
With recent the `MonadFail` changes, GHC now forces us to be honest that
these instances of `fail` are really just `error`.

### DIFF
--- a/src/Text/LLVM.hs
+++ b/src/Text/LLVM.hs
@@ -324,7 +324,7 @@ avoidName name = BB $ do
   rw <- get
   case avoid name (rwNames rw) of
     Just ns' -> set rw { rwNames = ns' }
-    Nothing  -> fail ("avoidName: " ++ name ++ " already registered")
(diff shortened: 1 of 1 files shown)


elliottt:
Could this be fixed by adding a `MonadFail` instance for `BB`?

atomb:
That would be a better long-term approach, but I think would require some more significant refactoring that would probably affect all users of the library (unless it was an instance that just said `fail = error`, which I'm not sure would really be an improvement). This change just fixes the build with 8.8 without changing behavior at all.

My inclination would be to merge this change as a clean compilation fix and then do some later work to refactor `BB` to allow recoverable error handling. It looks as though there are some existing calls to `error` in the `BB` monad that we could remove along the way.

elliottt:
That's a fair point. I'm happy to merge this and make an issue to follow up with the better long-term fix.

PR #74, opened 2020-08-17
Use Word32 or Word64 in LLVM types instead of Int32

robdockins:
For integer types, LLVM seems to only support up to 2^24 bits
anyway, but it doesn't make sense to allow them to be negative.
For Array and Vector types, LLVM uses uint_64 to describe
the number of elements, so we use Word64.

I guess this probably requires a version bump, not sure what version you want to go to.

### DIFF
--- a/src/Text/LLVM.hs
+++ b/src/Text/LLVM.hs
@@ -112,6 +112,7 @@ import Text.LLVM.AST
 import Control.Monad.Fix (MonadFix)
 import Data.Char (ord)
 import Data.Int (Int8,Int16,Int32,Int64)
+import Data.Word (Word32, Word64)
 import Data.Maybe (maybeToList)
 import Data.String (IsString(..))
 import MonadLib hiding (jump,Label)
@@ -413,7 +414,7 @@ terminateBasicBlock  = BB $ do
 
 -- Type Helpers ----------------------------------------------------------------
 
-iT :: Int32 -> Type
+iT :: Word32 -> Type
 iT  = PrimType . Integer
 
 ptrT :: Type -> Type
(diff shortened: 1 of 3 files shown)


elliottt (approved):
👍

elliottt:
We can bump to `0.11.0` whenever you're ready. Do you have any other changes you'd like to bundle into that release?

robdockins:
I don't have anything else urgent to go in right now, I think we can bump.

elliottt:
Sounds good, I'll bump it this evening.

robdockins:
Humm... should maybe hold off on a new release for the moment, I've just run some additional tests that are failing. I'll try to figure out what's going on.

robdockins:
Humm... some 30-bit and 10-bit values showing up in the DwarfVirtuality field.  Based on my limited looking around, this should be a quite small enum, so I'm not sure what's going on there.

robdockins:
Ugh... looks like LLVM changed the record format for DISubprogram at some point... all this logic is super gross:

https://github.com/llvm/llvm-project/blob/ef32c611aa214dea855364efd7ba451ec5ec3f74/llvm/lib/Bitcode/Reader/MetadataLoader.cpp#L1486

Looks like the relevant changes first appeared in LLVM 8

CF https://reviews.llvm.org/D54755

robdockins:
OK, sorry for the spam.  I'm now reasonably confident the issues I'm seeing are actually issues that need to be resolved inside the bitcode parser, and I don't think the AST nodes here need to change.

So, I think we're ready for a bump and release after all.

elliottt:
Good luck tracking down the DISubprogram changes :(

elliottt:
I've uploaded version [0.11.0](https://hackage.haskell.org/package/llvm-pretty-0.11.0) 👍

PR #76, opened 2021-03-16
Allow building with th-abstraction-0.4.*

RyanGlScott:
This is required to make `llvm-pretty` build with GHC 9.0, as only `th-abstraction-0.4` or later support 9.0.


elliottt:
Thank you!

PR #78, opened 2021-05-10
More visibility and linkage

### LINKED ISSUE #77: Define is missing the visibility field
Globals, functions and global aliases can have a visibility attribute, but in this package only globals have a visibility attribute.

noughtmare:
Fixes #77

I wasn't completely sure how to deal with the `declare` function. I could give it `FunAttrs` as argument like `define`, but `declare` doesn't have the garbage collection attribute. 

The main changes are:

   * add linkage and visibility to `Declare`
   * add visibility to `Define`
   * add linkage and visibility to `GlobalAlias`


elliottt:
This looks good to me, thanks for fixing this! I think it would be reasonable to introduce another type like `FunAttrs` for the `declare` function, if you'd like to tackle that in a follow-up PR.

PR #88, opened 2021-10-13
add LLVM12 poison value

Ptival:
Is anything else needed to add a simple value?


elliottt (approved):
Grepping through the llvm-12.x codebase, it does indeed look like this should work.

PR #90, opened 2021-12-03
Support structs with bitfields in `Text.LLVM.DebugUtils`

RyanGlScott:
LLVM bitcode doesn't directly record information about bitfields, but its debug information _does_ record this information. Knowing about bitfields is important for certain applications—see, for example, GaloisInc/saw-script#1461. This changes `Text.LLVM.DebugUtils` such that if any of the fields in a struct have bitfields, it will record this information in the new `BitfieldInfo` data type.

This requires a backwards-incompatible change to the type of the `Structure` data constructor. In case we need to add additional fields to `Structure` in the future, I converted `Structure`'s fields into a record data type, which makes it slightly easier to extend. I also did the same thing to `Union` for consistency (although this is not strictly necessary).

### DIFF
--- a/src/Text/LLVM/DebugUtils.hs
+++ b/src/Text/LLVM/DebugUtils.hs
@@ -10,7 +10,8 @@ Point-of-contact : emertens
 -}
 module Text.LLVM.DebugUtils
   ( -- * Definition type analyzer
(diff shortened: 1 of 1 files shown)


elliottt (approved):
This seems super reasonable, thanks for adding the thorough comments!

PR #91, opened 2022-02-11
Tweaks to the DebugUtils module

robdockins:
These changes expose a bit more information gleaned from the debug
information about types from the source program.

This changes the type of `computeFunctionTypes` to be a bit more accurate, I think.  I don't know of any use sites for this function, so I don't know how disruptive this will be.

### DIFF
--- a/src/Text/LLVM/DebugUtils.hs
+++ b/src/Text/LLVM/DebugUtils.hs
@@ -24,6 +24,7 @@ module Text.LLVM.DebugUtils
 
   -- * Info hueristics
   , guessAliasInfo
+  , guessTypeInfo
   ) where
 
 import           Control.Applicative    ((<|>))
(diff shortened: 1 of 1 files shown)


src/Text/LLVM/DebugUtils.hs
```diff
@@ -268,9 +271,9 @@ debugInfoToUnionField mdMap di =
 computeFunctionTypes ::
   Module       {- ^ module to search                     -} ->
   Symbol       {- ^ function symbol                      -} ->
-  Maybe [Info] {- ^ return and argument type information -}
+  Maybe [Maybe Info] {- ^ return and argument type information -}
 computeFunctionTypes m sym =
-  [ maybe (BaseType "void") (valMdToInfo mdMap) <$> types
```
elliottt, line 273: Does this change mean that `void` is now represented as `Nothing`?

elliottt (approved):
I think this makes sense. Should we bump llvm-pretty hackage soon?

src/Text/LLVM/DebugUtils.hs
robdockins: Right.  I believe that is more faithful to the DWARF encoding.

robdockins:
> I think this makes sense. Should we bump llvm-pretty hackage soon?

Perhaps? I'm not sure how much has accumulated since the last release.

PR #93, opened 2022-04-29
Add callbr instruction

langston-barrett:
This was added in LLVM 9 in commit llvm/llvm-project@784929d.

Also fix bug around printing of blockaddress constants which appear as operands to callbrs.

Bitcode parser PR: https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/187

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -764,6 +764,17 @@ data Instr' lab
          * Middle of basic block.
          * The result is as indicated by the provided type. -}
 
+  | CallBr Type (Value' lab) [Typed (Value' lab)] lab [lab]
+    {- ^ * Call a function in asm-goto style:
+             return type;
+             function operand;
+             arguments;
+             default basic block destination;
+             other basic block destinations.
+         * Middle of basic block.
+         * The result is as indicated by the provided type.
+         * Introduced in LLVM 9. -}
+
   | Alloca Type (Maybe (Typed (Value' lab))) (Maybe Int)
     {- ^ * Allocated space on the stack:
            type of elements;
@@ -943,6 +954,7 @@ isTerminator instr = case instr of
   Ret{}        -> True
   RetVoid      -> True
   Jump{}       -> True
+  CallBr{}     -> True
   Br{}         -> True
   Unreachable  -> True
   Unwind       -> True
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -506,6 +506,7 @@ ppInstr instr = case instr of
   Conv op a ty           -> ppConvOp op <+> ppTyped ppValue a
                         <+> "to" <+> ppType ty
   Call tc ty f args      -> ppCall tc ty f args
+  CallBr ty f args u es  -> ppCallBr ty f args u es
   Alloca ty len align    -> ppAlloca ty len align
   Load ptr mo ma         -> ppLoad ptr mo ma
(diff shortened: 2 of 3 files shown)


langston-barrett:
Would also appreciate a review from @RyanGlScott

elliottt (approved):
This looks good to me! I've invited @RyanGlScott as a collaborator so that they can approve as well.

src/Text/LLVM/AST.hs
```diff
@@ -764,6 +764,16 @@ data Instr' lab
          * Middle of basic block.
          * The result is as indicated by the provided type. -}
 
+  | CallBr Type (Value' lab) [Typed (Value' lab)] lab [lab]
+    {- ^ * Call a function in asm-goto style:
```
RyanGlScott, line 768: It would be worth mentioning that this was introduced in LLVM 9.0.

src/Text/LLVM/PP.hs
```diff
@@ -667,6 +668,18 @@ ppCall tc ty f args
   body = "call" <+> ppCallSym ty f
       <> parens (commas (map (ppTyped ppValue) args))
 
+ppCallBr :: LLVM => Type -> Value -> [Typed Value] -> BlockLabel -> [BlockLabel] -> Doc
+ppCallBr ty f args to indirectDests =
+  "callbr"
+     <+> ppType res <+> ppValue f <> parens (commas (map (ppTyped ppValue) args))
+     <+> "to" <+> ppLab to <+> brackets (commas (map ppLab indirectDests))
```
RyanGlScott, line 677: The example [here](https://github.com/llvm/llvm-project/commit/784929d0454c4df6a98ef6fbbd1d30a6f71f9c16#diff-1d1772820f1edf0653335111c7ab690c53a2df20887eadadf55d7a9c17aa907aR6851-R6852) includes an `or jump [labels]` bit. Does this code print out the `or jump` part? (`llvm-pretty-bc-parser`'s CI would catch this, I think, but it hasn't run on GaloisInc/llvm-pretty-bc-parser#187 yet.)

langston-barrett: Hmm, interesting. The current Language Reference doesn't have that syntax: https://llvm.org/docs/LangRef.html#callbr-instruction.
```
<result> = callbr [cconv] [ret attrs] [addrspace(<num>)] <ty>|<fnty> <fnptrval>(<function args>) [fn attrs]
              [operand bundles] to label <fallthrough label> [indirect labels]
```
I also did run the disassembly tests on the testcase in the bitcode parser PR and they passed, but that was probably with whatever version of LLVM I have installed and got run in the tests.

RyanGlScott: Ah, good catch. The link that I gave above is the syntax used in LLVM 9.0, but it changed in LLVM 10.0 in llvm/llvm-project@93d87260f1f7b5ec6775fe88891e47a57e6392d6. I'm not quite sure if there is a good way to handle the difference in LLVM versions here, but I'm definitely more inclined to pick the modern syntax.

This might be worth leaving a comment about.

RyanGlScott (approved):

src/Text/LLVM/PP.hs
langston-barrett: Done.

langston-barrett: Here are the disasm tests passing on the bitcode parser in CI, for good measure: https://github.com/GaloisInc/llvm-pretty-bc-parser/runs/6233698029?check_suite_focus=true#step:7:29

PR #94, opened 2022-05-13
Upstream debug utilities from Crucible, SAW

langston-barrett:
Upstream a few utilities. At least one of these was duplicated (copy/pasted) between the two codebases, so seems worth it to collect them together. See https://github.com/GaloisInc/crucible/issues/696.

### DIFF
--- a/src/Text/LLVM/DebugUtils.hs
+++ b/src/Text/LLVM/DebugUtils.hs
@@ -25,6 +25,9 @@ module Text.LLVM.DebugUtils
   -- * Info hueristics
   , guessAliasInfo
   , guessTypeInfo
+
+  -- * Function arguments
+  , debugInfoArgNames
   ) where
 
 import           Control.Applicative    ((<|>))
@@ -432,3 +435,34 @@ guessTypeInfo mdMap name =
           = Just (debugInfoToInfo mdMap di)
 
     go _ = Nothing
+
+------------------------------------------------------------------------
+
+-- | Find source-level names of function arguments
+debugInfoArgNames :: Module -> Define -> IntMap String
+debugInfoArgNames m d =
+  case Map.lookup "dbg" $ defMetadata d of
+    Just (ValMdRef s) -> scopeArgs s
+    _ -> IntMap.empty
+  where
+    scopeArgs :: Int -> IntMap String
+    scopeArgs s = go $ modUnnamedMd m
+      where
+        go :: [UnnamedMd] -> IntMap String
+        go [] = IntMap.empty
+        go
+          ( UnnamedMd
+              { umValues =
+                  ValMdDebugInfo
+                    ( DebugInfoLocalVariable
+                        DILocalVariable
+                          { dilvScope = Just (ValMdRef s'),
+                            dilvArg = a,
+                            dilvName = Just n
+                          }
+                      )
+              }
+              : xs
+            ) =
+            if s == s' then IntMap.insert (fromIntegral a - 1) n $ go xs else go xs
+        go (_ : xs) = go xs
--- a/src/Text/LLVM/DebugUtils.hs
+++ b/src/Text/LLVM/DebugUtils.hs
@@ -1,12 +1,12 @@
 {-# Language TransformListComp, MonadComprehensions #-}
 {- |
-Module           : $Header$
+Module           : Text.LLVM.DebugUtils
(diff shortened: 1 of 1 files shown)


src/Text/LLVM/DebugUtils.hs
```diff
@@ -447,3 +439,75 @@ guessTypeInfo mdMap name =
           = Just (debugInfoToInfo mdMap di)
 
     go _ = Nothing
+
+------------------------------------------------------------------------
+
+-- | Find source-level names of function arguments
+debugInfoArgNames :: Module -> Define -> IntMap String
+debugInfoArgNames m d =
+  case Map.lookup "dbg" $ defMetadata d of
```
RyanGlScott: We might as well use the conventions in this module:

```suggestion
  case Map.lookup dbgKind $ defMetadata d of
```

src/Text/LLVM/DebugUtils.hs
```diff
@@ -447,3 +439,75 @@ guessTypeInfo mdMap name =
           = Just (debugInfoToInfo mdMap di)
 
     go _ = Nothing
+
+------------------------------------------------------------------------
+
+-- | Find source-level names of function arguments
+debugInfoArgNames :: Module -> Define -> IntMap String
```
RyanGlScott, line 446: I wonder if it would make more sense to use `Ident` instead of `String` here for symmetry with `localVariableNameDeclarations`. Similarly for `debugInfoGlobalLines`' and `debugInfoDefineLines`' use of `String`.

src/Text/LLVM/DebugUtils.hs
```diff
@@ -447,3 +439,75 @@ guessTypeInfo mdMap name =
...
+              : xs
+            ) =
+            if s == s' then IntMap.insert (fromIntegral a - 1) n $ go xs else go xs
+        go (_ : xs) = go xs
+
+------------------------------------------------------------------------
+
+-- Helper, not exported
+mapFold :: Ord k => (a -> Maybe (k, v)) -> [a] -> Map k v
+mapFold f = foldr go Map.empty
```
RyanGlScott: If I were writing this, I'd implement it as simply `mapFold f = Map.fromList . mapMaybe f`. Your call as to which one you like better.

langston-barrett: Good point, and at that point there's not much point having it un-inlined. I also updated the above function using `IntMap` for consistency.

RyanGlScott (approved):

src/Text/LLVM/DebugUtils.hs
langston-barrett: After thinking more about this, I'm not so sure. `Ident` is used for:

* [names of types](https://github.com/elliottt/llvm-pretty/blob/d2fa59fc2e2f32df3d720ed3bde41924a0ec558e/src/Text/LLVM/AST.hs#L390), 
* [names of function arguments](https://github.com/elliottt/llvm-pretty/blob/d2fa59fc2e2f32df3d720ed3bde41924a0ec558e/src/Text/LLVM/AST.hs#L448), 
* [names of blocks](https://github.com/elliottt/llvm-pretty/blob/d2fa59fc2e2f32df3d720ed3bde41924a0ec558e/src/Text/LLVM/AST.hs#L502), and 
* [names of variables (instruction results)](https://github.com/elliottt/llvm-pretty/blob/d2fa59fc2e2f32df3d720ed3bde41924a0ec558e/src/Text/LLVM/AST.hs#L1065).

For `debugInfoArgNames`, the values are source-level identifiers, so they have no correspondence with LLVM identifiers. So we might want to avoid `Ident`, since we're not using it as it is already used in the AST.

For `debugInfoGlobalLines` and `debugInfoDefineLines`, it might make more sense to use a `Symbol`. However, it's also not clear to me that [`DIGlobalVariable`](https://llvm.org/docs/LangRef.html#diglobalvariable) and [`DISubprogram`](https://llvm.org/docs/LangRef.html#disubprogram) are guaranteed to have names that actually correspond to the definition's name in the LLVM module, so perhaps it would be best to avoid `Symbol` as well to avoid insinuating that they do. Thoughts @RyanGlScott?

RyanGlScott: You've convinced me. In that case, I'm fine with sticking with the types as-is.

PR #95, opened 2022-09-06
Fix printing error for FP80 values.

kquick:


elliottt:
This seems reasonable, but could you include an example of the existing behavior and the corrected new behavior?

kquick:
The old version was just printing the remaining bytes after the shift, yielding:
```
  %173 = fmul x86_fp80 %172, 0xK4040080111f11f4011f400011f40000011f4000000011f400000000011f400000000000
```
and with the newer version, it matches `llvm-dis`, properly printing:
```
  %173 = fmul x86_fp80 %172, 0xK4008011f400000000000
```

elliottt (approved):

PR #96, opened 2022-09-23
Parsing target triples

### LINKED ISSUE #21: Parse/print target triples
Does llvm-pretty have the ability to represent target triples?  A glance through the AST doesn't show any obvious place where that information is captured.  It would be nice to be able to parse and print the target triple information.  In particular, I'd like to be able to know the architecture the code was compiled for when reading bit code via llvm-pretty-bc-parser.

langston-barrett:
Fixes #21.

### DIFF
--- a/src/Text/LLVM/Triple/AST.hs
+++ b/src/Text/LLVM/Triple/AST.hs
@@ -428,6 +428,9 @@ instance Monoid ObjectFormat where
   mempty = UnknownObjectFormat
 
 -- | More like a sextuple than a triple, if you think about it.
+--
+-- The LLVM version of this holds onto the un-normalized string representation.
+-- We discard it.
 data TargetTriple
   = TargetTriple
     { ttArch :: Arch
--- a/src/Text/LLVM/Triple/AST.hs
+++ b/src/Text/LLVM/Triple/AST.hs
@@ -1,0 +1,464 @@
+{- |
+Module      : Text.LLVM.Triple.AST
+Description : AST of LLVM target triples.
+License     : BSD3
+Maintainer  : Langston Barrett
+Stability   : experimental
+-}
+
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE StrictData #-}
+
+module Text.LLVM.Triple.AST
+  ( Arch(..)
+  , SubArch(..)
+  , Vendor(..)
+  , OS(..)
+  , Environment(..)
+  , ObjectFormat(..)
+  , TargetTriple(..)
+  ) where
+
+import Data.Data (Data)
+import Data.Typeable (Typeable)
+import GHC.Generics (Generic)
+
+-- | The constructors of this type exactly mirror the LLVM @enum ArchType@,
+-- including inconsistent labeling of endianness. Capitalization is taken from
+-- the LLVM comments, which are reproduced below.
+--
+-- Last updated: LLVM 15.0.1:
+-- https://github.com/llvm/llvm-project/blob/llvmorg-15.0.1/llvm/include/llvm/ADT/Triple.h#L46
+data Arch
+  = UnknownArch
+    -- | ARM (little endian): arm armv.* xscale
+  | ARM
+    -- | ARM (big endian): armeb
+  | ARMEB
+    -- | AArch64 (little endian): aarch64
+  | AArch64
+    -- | AArch64 (big endian): aarch64_be
+  | AArch64_BE
+    -- | AArch64 (little endian) ILP32: aarch64_32
+  | AArch64_32
+    -- | ARC: Synopsys ARC
+  | ARC
+    -- | AVR: Atmel AVR microcontroller
+  | AVR
+    -- | eBPF or extended BPF or 64-bit BPF (little endian)
+  | BPFEL
+    -- | eBPF or extended BPF or 64-bit BPF (big endian)
+  | BPFEB
+    -- | CSKY: csky
+  | CSKY
+    -- | DXIL 32-bit DirectX bytecode
+  | DXIL
+    -- | Hexagon: hexagon
+  | Hexagon
+    -- | LoongArch (32-bit): loongarch32
+  | LoongArch32
+    -- | LoongArch (64-bit): loongarch64
+  | LoongArch64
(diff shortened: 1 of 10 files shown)


src/Text/LLVM/Triple/Parse/ARM.hs
```diff
@@ -0,0 +1,348 @@
...
+--------------------------------------------------------------------------------
+-- Arch
+
+-- | @llvm::ARM::getArchSynonym@
+getArchSynonym :: ArchName -> ArchName
+getArchSynonym (ArchName arch) =
+  ArchName $
+    if | cases ["v5"] -> "v5t"
+       | cases ["v5e"] -> "v5te"
+       | cases ["v5e"] -> "v5te"
```
RyanGlScott: The second case here is redundant, unless I'm missing something.

RyanGlScott:
Is it expected that this fails to parse the ELF object format?

```
λ> parseTriple "i386-unknown-linux-elf"
TargetTriple {ttArch = X86, ttSubArch = NoSubArch, ttVendor = UnknownVendor, ttOS = Linux, ttEnv = UnknownEnvironment, ttObjFmt = UnknownObjectFormat}
```

src/Text/LLVM/Triple/Parse/ARM.hs
langston-barrett: Good catch, thanks!

langston-barrett:
> Is it expected that this fails to parse the ELF object format?

Indeed not. The problem is that I was treating object formats as a *fifth* field of a target triple, when they apparently *overlap* with the "environment" as the *fourth* field. Thanks for pointing this out, I've added a test with that triple!

src/Text/LLVM/Triple/Parse.hs
```diff
@@ -0,0 +1,294 @@
...
+                | armArch == ARM.ARMV8_1MMainline -> ARMSubArch_v8_1m_mainline
+                | otherwise -> NoSubArch
+  where
+    startsWith = (`List.isPrefixOf` subArchName)
+    endsWith = (`List.isSuffixOf` subArchName)
+    armSubArch = ARM.getCanonicalArchName (ARM.ArchName subArchName)
+
+-- | @llvm::Triple::getDefaultFormat@
+--
+-- TODO(lb): Implement me!
```
RyanGlScott: Is this something you plan to do in this PR?

src/Text/LLVM/Triple/Parse/ARM.hs
```diff
@@ -0,0 +1,347 @@
...
+  in case mArchName of
+      Nothing -> UnknownArch
+      Just (ArchName archNm) ->
+        if | isa == Just ISAThumb &&
+             ("v2" `List.isPrefixOf` archNm || "v3" `List.isPrefixOf` archNm) -> UnknownArch
+            -- TODO(lb): LLVM has one more check here involving the "arch
+            -- profile" that's not yet implemented here... Probably not a big
+            -- deal because this case is only executed when the target arch
+            -- doesn't match the "canonical" versions like "arm", "thumb", and
+            -- "aarch64".
```
RyanGlScott: Similar question here.

src/Text/LLVM/Triple/AST.hs
```diff
@@ -0,0 +1,467 @@
...
+License     : BSD3
+Maintainer  : Langston Barrett
+Stability   : experimental
+-}
+
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE StrictData #-}
+
+module Text.LLVM.Triple.AST
```
RyanGlScott, line 13: Do you think it is worth reexporting this module from `Text.LLVM.AST`?

RyanGlScott (approved):

src/Text/LLVM/Triple/AST.hs
langston-barrett: > Do you think it is worth reexporting this module from Text.LLVM.AST?

Quite frankly, I don't have any heuristics at all about when it is good to re-export things. I'm open to suggestions, and would especially appreciate any general guidance you have on the subject!

src/Text/LLVM/Triple/Parse.hs
langston-barrett: I could, but I don't think it's important for my use-case (Crucible). We should make an issue and replace `lb` with the number if we decide to defer.

src/Text/LLVM/Triple/Parse/ARM.hs
langston-barrett: Again, it's likely not necessary for my use-case but I can implement it if there's a reason to do so.

src/Text/LLVM/Triple/Parse.hs
RyanGlScott: OK, I am alright with opening an issue about this and deferring it.

src/Text/LLVM/Triple/AST.hs
RyanGlScott: I don't have anything resembling heuristics, only intuition. I bring this up since it's a bit odd that `Text.LLVM.AST` now exports all of the AST types _except_ the target triple–related ones, which intuitively feels a bit strange. Ultimately, it's little more than an intuition, so I'll leave it up to you to decide whether it's worth doing or not.

elliottt (approved):
This looks great to me, thank you for adding this! When you do create the issues, could you link them back in the comment threads?

src/Text/LLVM/Triple/Parse.hs
langston-barrett: https://github.com/elliottt/llvm-pretty/issues/97

src/Text/LLVM/Triple/Parse/ARM.hs
langston-barrett: https://github.com/elliottt/llvm-pretty/issues/98

src/Text/LLVM/Triple/AST.hs
langston-barrett: I'm going to leave this off for now because `Text.LLVM.AST` lacks an explicit export list, so it would be a bit of a pain.

langston-barrett:
> When you do create the issues, could you link them back in the comment threads?

For sure, done!

PR #99, opened 2022-09-27
Run cabal check in CI

langston-barrett:
Since llvm-pretty is published to Hackage, we should ensure that it doesn't accumulate changes that would cause updates to be rejected. cabal check is very fast, so this doesn't add much time to CI builds.


elliottt (approved):
:shipit:

Issue #102, opened 2022-09-27
Support opaque pointers


langston-barrett:
LLVM is in the process of migrating all of its pointer types from `i8*`, `i32*`, etc. to an opaque `ptr` type, as described [here](https://llvm.org/docs/OpaquePointers.html). LLVM 13 takes an important step in that direction, as it is the first LLVM release to include `ptr` in the LLVM AST. See https://github.com/llvm/llvm-project/commit/2155dc51d700c9fb5f29d79eaacf5e1470e4d8ca.

We should support opaque pointers in the LLVM AST and pretty-printer. In particular, we will have to either add a new `OpaquePointer` constructor that doesn't take a pointee type as an argument, or make [`PtrTo`](https://github.com/elliottt/llvm-pretty/blob/31612ff381fba3971263371daab452faab5b76fe/src/Text/LLVM/AST.hs#L262) take a `Maybe`.

See also: https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/177

RyanGlScott:
Hah, I just learned that LLVM also has an [`opaque`](https://releases.llvm.org/15.0.0/docs/LangRef.html#opaque-structure-types) structure type that is completely unrelated to opaque pointers (just in case you thought there weren't enough concepts with "opaque" in the name), which `llvm-pretty` represents as [`Opaque`](https://github.com/elliottt/llvm-pretty/blob/31612ff381fba3971263371daab452faab5b76fe/src/Text/LLVM/AST.hs#L266). I wonder if it would be confusing to have both `Opaque` and `OpaquePointer`.

RyanGlScott:
To echo similar sentiments as in https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/177#issuecomment-1476070897, I've discovered that adding support the opaque `ptr` type in `llvm-pretty` itself is actually quite straightforward. The tricky part is making sure that instructions which _use_ opaque pointers behave as expected.

[...]

PR #104, opened 2023-03-21
Basic support for opaque pointers

RyanGlScott:
This is a collection of two patches that make it possible for `llvm-pretty` to handle opaque pointers and pretty-print them correctly when used in a `call` instruction, the latter of which was prompted by https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/189#issuecomment-1476068137.

This is likely not everything that we will need to handle everything opaque pointer–related, but it is a good starting point. See #102 for the long-term goal of robustly supporting opaque pointers.


langston-barrett (approved):
Love the docs!

elliottt (approved):
Looks great to me, +1 on the docs!

PR #109, opened 2023-04-06
Support metadata fields added in LLVM 14

RyanGlScott:
The primary goal of this PR is to add support for `btf_tag` annotations on LLVM's metadata nodes, which are necessary to load essentially any bitcode file compiled with `clang-14 -g`. See https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/188.

This also adds support for DWARF address spaces in `DIDerivedType` (see https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/85) and alignment in `DILocalVariable` (see https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/217), both of which appear immediately before `btf_tag` annotations in their respective metadata nodes. These do not occur as often as `btf_tag` annotations do, but I added them for the sake of completeness.

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -1271,6 +1271,8 @@ data DICompositeType' lab = DICompositeType
   , dictAssociated     :: Maybe (ValMd' lab)
   , dictAllocated      :: Maybe (ValMd' lab)
   , dictRank           :: Maybe (ValMd' lab)
+  , dictAnnotations    :: Maybe (ValMd' lab)
+    -- ^ Introduced in LLVM 14.
   } deriving (Data, Eq, Functor, Generic, Generic1, Ord, Show, Typeable)
 
 type DICompositeType = DICompositeType' BlockLabel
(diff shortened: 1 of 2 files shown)


src/Text/LLVM/AST.hs
```diff
@@ -1271,22 +1271,31 @@ data DICompositeType' lab = DICompositeType
...
-  , didtName :: Maybe String
-  , didtFile :: Maybe (ValMd' lab)
-  , didtLine :: Word32
-  , didtScope :: Maybe (ValMd' lab)
-  , didtBaseType :: Maybe (ValMd' lab)
-  , didtSize :: Word64
-  , didtAlign :: Word64
-  , didtOffset :: Word64
-  , didtFlags :: DIFlags
-  , didtExtraData :: Maybe (ValMd' lab)
```
langston-barrett, line 1289: Generally, I prefer non-aligned type ascriptions for the reason that, as in this PR, adding a longer field name causes a diff that includes *all* of the fields, making it harder to distinguish what's changed. That being said, it looks like this changed to be more consistent with other parts of the file, which is also important.

Also looks like maybe I did this? :sweat_smile: So I can't complain too much.

langston-barrett (approved):

src/Text/LLVM/AST.hs
RyanGlScott: Indeed, this is a cherry-picked version of https://github.com/elliottt/llvm-pretty/pull/37, so I blame you for this :)

Then again, I have already performed some surgery on that commit, so I would be fine with removing this particular part of the commit (or factoring it out into a separate commit).

langston-barrett: I'm fine with whatever, not a big deal either way!

RyanGlScott: I opted to just revert the other alignment-related changes, as this file never consistently uses one convention for all its data type declarations.

PR #110, opened 2023-04-21
More complete support for opaque pointers (required for LLVM 15+)

### LINKED ISSUE #102: Support opaque pointers

RyanGlScott:
This is a collection of various tweaks and additions to the `llvm-pretty` API that allow it to fully support opaque pointers, an alternative representation of pointer types that is enabled by default in LLVM 15 or later. This patch should fix #102 insofar as I have tested these changes on a reasonable subset of LLVM 15 bitcode, and this is enough to make everything that I have thrown at `llvm-pretty` work.

I have split this patch up into five self-contained commits, and for reviewing purposes, I'd recommend viewing each patch in isolation. The general themes of these patches are:

## Pretty-printing fixes

* `Fix CallBr pretty-printing w.r.t. opaque pointers`: This is a minor change to `llvm-pretty`'s pretty-printer to make it print out the function type stored directly in the `CallBr` data constructor, rather than trying to inspect the pointee type of the pointer argument (which will not exist if an opaque pointer is used).

## Storing types directly in the AST

These are the most significant changes, as they require changing the types of `Instr'` data constructors:

* `Store load type separately in Load`: We now store the load type directly rather than trying to infer it from a pointee type.
* `Store basis type separately in GEP/ConstGEP`: We now store the type used as the basis for a `getelementptr` calculation directly rather than trying to infer it from a pointee type. In the case of `ConstGEP`, this basis type corresponds to the first argument, which I have made an explicit field in `ConstGEP` for symmetry with `GEP`.

## Utilities for inspecting and manipulating opaque pointers

These commits add opaque pointer–related utilities that are essential to supporting `llvm-pretty`'s approach of allowing both opaque and non-opaque pointers.

* `Provide ability to check type equality modulo opaque pointers`: This implements `eqTypeModuloOpaquePtrs` and `cmpTypeModuloOpaquePtrs` functions that compare `Type`s, but treat `PtrOpaque` as being identical to `PtrTo`. This is extremely useful in libraries like `llvm-pretty-bc-parser` and `crucible-llvm`, which often use this notion of `Type` equality.
* `Define fixupOpaquePtrs function to support mixing opaque/non-opaque pointers`: This implements a `fixupOpaquePtrs` function that takes an `llvm-pretty` AST and, if it contains at least one opaque pointer, converts all non-opaque pointers to opaque ones. This is useful for pretty-printing purposes, as tools like `llvm-as` do not permit mixing opaque and non-opaque pointers, so this function makes it possible to still interface with these tools.

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -814,8 +814,9 @@ data Instr' lab
          * Middle of basic block.
          * Returns a pointer to hold the given number of elements. -}
 
-  | Load (Typed (Value' lab)) (Maybe AtomicOrdering) (Maybe Align)
+  | Load Type (Typed (Value' lab)) (Maybe AtomicOrdering) (Maybe Align)
     {- ^ * Read a value from the given address:
+           type being loaded;
            address to read from;
            atomic ordering;
            assumptions about alignment of the given pointer.
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -12,8 +12,10 @@ import Data.Data (Data)
 import Data.Typeable (Typeable)
 import Control.Monad (MonadPlus(mzero,mplus),(<=<),guard)
 import Data.Int (Int32,Int64)
+import Data.Generics (everywhere, extQ, mkT, something)
 import Data.List (genericIndex,genericLength)
 import qualified Data.Map as Map
+import Data.Maybe (isJust)
 import Data.Semigroup as Sem
 import Data.String (IsString(fromString))
 import Data.Word (Word8,Word16,Word32,Word64)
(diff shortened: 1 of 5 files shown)


langston-barrett (approved):
Thanks for splitting this into such nice, focused commits! Very easy to review.

kquick (approved):

elliottt (approved):
This is great! Thanks for keeping llvm-pretty up to date :D

src/Text/LLVM/AST.hs
```diff
@@ -347,6 +349,98 @@ isPointer PtrOpaque = True
...
+
+-- | Like `Type'`, but where the 'PtrTo' and 'PtrOpaque' constructors have been
+-- collapsed into a single 'PtrView' constructor. This provides a coarser notion
+-- of type equality than what `Type'` provides, which distinguishes the two
+-- types of pointers.
+--
+-- `TypeView'` is not used directly in any of the other AST types. Instead, it
+-- is primarly used as an auxiliary type to power the 'eqTypeModuloOpaquePtrs'
+-- and 'cmpTypeModuloOpaquePtrs' functions.
+data TypeView' ident
```
RyanGlScott, line 515: It was pointed out to me that this patch exports the `TypeView'` data type by virtue of `Text.LLVM.AST` not having an explicit export list. This isn't what I intended, as `TypeView'` is meant to be internal to `eqTypeModuloOpaquePtrs` and `cmpTypeModuloOpaquePtrs` functions. We should add an explicit export list to `Text.LLVM.AST` to make sure that this actually happens.

PR #115, opened 2023-07-10
Fix some minor pretty-printing bugs

### LINKED ISSUE #113: `dicuRangesBaseAddress` should be a `Maybe Bool` instead of a `Bool` to support old LLVMs
Currently, [`dicuRangesBaseAddress`](https://github.com/elliottt/llvm-pretty/blob/94e384842b214ba72446d1694446fb5261ab6ce2/src/Text/LLVM/AST.hs#L1498) is of type `Bool`, and it will always be printed out when pretty-printing an LLVM module. But `rangesBaseAddress` was only introduced in LLVM 11, and older LLVMs are incapable of parsing `rangesBaseAddress: false` in the pretty-printed output:

```
$ llvm-as-10 /tmp/smallprogllvm-disasm1951375-146.ll
<snip>
llvm-as: /tmp/smallprogllvm-disasm1951375-146.ll:228:43: error: invalid field 'rangesBaseAddress'
                        nameTableKind: 2, rangesBaseAddress: false)
```

We should match what LLVM does [here](https://github.com/llvm/llvm-project/blob/9841daf27076886c6ab8e155eb812bda76f77532/llvm/lib/AsmParser/LLParser.cpp#L5215) and make `rangesBaseAddress` optional such that if a `rangesBaseAddress` value isn't specified, then it isn't printed out. Concretely, that would involve changing the type of `dicuRangesBaseAddress` from `Bool` to `Maybe Bool`.

### LINKED ISSUE #114: Label targets are not represented correctly for llvm-as consumption
Bitcode generated by clang and then parsed via llvm-pretty-bc-parser and then output via this library into `.ll` files may not be usable with `llvm-as` because of the representation of the label targets.

Generated code contains `call void @llvm.db.label(metadata !DILabel(scope: !8, name: loop,` which fails on `loop` (presumably because it needs to be quoted).

This error can be seen running `cabal run disasm-test -- -p '/callbr.c/' (requires https://github.com/galoisinc/llvm-pretty-bc-parser/pulls/230).

RyanGlScott:
This:

* Escapes the names of `DILabel`s in double quotes. This fixes #114.
* Makes `dicuRangesBaseAddress :: Maybe Bool` and omits pretty-printing it if is has the value `Nothing`. This fixes #113.

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -1495,7 +1495,7 @@ data DICompileUnit' lab = DICompileUnit
   , dicuSplitDebugInlining :: Bool
   , dicuDebugInfoForProf   :: Bool
   , dicuNameTableKind      :: Word64
-  , dicuRangesBaseAddress  :: Bool
+  , dicuRangesBaseAddress  :: Maybe Bool
   , dicuSysRoot            :: Maybe String
   , dicuSDK                :: Maybe String
   } deriving (Data, Eq, Functor, Generic, Generic1, Ord, Show, Typeable)
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -1008,7 +1008,7 @@ ppDICompileUnit' pp cu = "!DICompileUnit"
        , pure ("splitDebugInlining:"    <+> ppBool (dicuSplitDebugInlining cu))
        , pure ("debugInfoForProfiling:" <+> ppBool (dicuDebugInfoForProf cu))
        , pure ("nameTableKind:"         <+> integral (dicuNameTableKind cu))
-       , pure ("rangesBaseAddress:"     <+> ppBool (dicuRangesBaseAddress cu))
+       ,     (("rangesBaseAddress:"     <+>) . ppBool) <$> (dicuRangesBaseAddress cu)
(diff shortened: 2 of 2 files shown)


elliottt:
The changes seem good to me, but I'm happy to re-review when this is marked ready :)

RyanGlScott:
For context: these are both issues that we ran into while trying to make the test case from https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/223 work on older LLVMs. I was hoping that these two commits would be enough to make that test case go through, but sadly, we uncovered yet _another_ issue in #116, and fixing #116 is unlikely to be as straightforward.

Considering that this PR does introduce a breaking API change, I'm inclined to park this until we know for sure that this approach will in fact fix https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/223, #116, and any related issues we encounter in the process.

RyanGlScott:
This PR fixes the issue by wrapping a field in a `Maybe`, which is a rather heavy-handed fix. A better solution would be to change the behavior of the pretty-printer depending on whether LLVM 11+ is being used or not. And in fact, @kquick's recent series of PRs have now introduced machinery to make this possible, and @kquick proposes fixing the issue by using this machinery. I'm inclined to agree, so I'll close this PR in favor of this alternative solution.

RyanGlScott:
Superseded by #121.

PR #117, opened 2023-08-03
Add output tests

kquick:
Adds some simple pretty-printing output verification tests. This amounts to spot-checking a few places in the pretty printing that are examined by hand: in general, it its much more effective to do round-trip testing using AST's parsed from bitcode generated by actual programs, which is what occurs in the `llvm-pretty-bc-parser` package (i.e. much more comprehensive testing is deferred to the `llvm-pretty-bc-parser` package, which might then reveal issues that need to be fixed in this `llvm-pretty` package).


elliottt:
This is great, thank you!

elliottt (approved):

.gitignore
```diff
@@ -1,4 +1,11 @@
...
 .cabal-sandbox/
 cabal.sandbox.config
-dist/
-dist-newstyle/
+.stack-work/
+.ghc.environment.*
+/dist-newstyle
+/fuzz-temp-test.c
+/fuzz-temp-test.bc
+/fuzz-results
```
RyanGlScott: What generates these? Worth leaving a comment.

test/Output.hs
```diff
@@ -0,0 +1,223 @@
...
+tests :: Tasty.TestTree
+tests = Tasty.testGroup "LLVM pretty-printing output tests"
+  $ let -- s1 is a non-sensical construct whose primary intention is to hold two
+        -- sub-structures that change their pretty representations at different
+        -- LLVM versions.  The pretty output will be checked at different LLVM
+        -- versions to ensure that the desired version-specific changes in the
+        -- output are seen.
+        s1 = Effect
+             (GEP True (Alias (Ident "hi")) (Typed Opaque dcu) [])
+             [] :: Stmt
```
RyanGlScott: Nit pick: I'd prefer putting the type signature up front:

```hs
s1 :: Stmt
s1 = Effect
     (GEP ...)
     []
```

test/Output.hs
```diff
@@ -0,0 +1,223 @@
...
+      --------
+      |]
+
+  ]
+
+----------------------------------------------------------------------
+
+assertEqLines :: T.Text -> T.Text -> IO ()
+assertEqLines t1 t2 =
+  if t1 == t2 then return () else assertFailure $ multiLineDiff t1 t2
```
RyanGlScott:
```suggestion
  unless (t1 == t2) $ assertFailure $ multiLineDiff t1 t2
```

RyanGlScott (approved):
I like it. Some minor suggestions inline.

.gitignore
kquick: Oops, those were in the copy paste from llvm-pretty-bc-parser and aren't actually needed here.  Removed.

test/Output.hs
kquick: Updated.

test/Output.hs
kquick: Good call.  Updated.

kquick:
I do not have merge privileges, so if this looks good, could one of you go ahead and merge this, please.

PR #118, opened 2023-08-04
Use Fmt encapsulation and add LLVMPretty class

kquick:
Internally, the pretty-printing now uses a type alias `Fmt` to represent the operation that converts something to a `Doc`, with the presence of the `Config` implicit parameter.  This allows standardization of the printing operation declarations as a precursor to making changes that affect all operations.

This also adds an `LLVMPretty` class so that the generic `llvmPP` method can be used to perform type-based dispatching to the appropriate pretty-printer.  Currently, only the most prominently used instances are defined, but more will be coming in future changes.

### DIFF
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -34,8 +34,6 @@ import Prelude hiding ((<>))
 
 -- Pretty-printer Config -------------------------------------------------------
 
-type LLVM = ?config :: Config
-
 -- | The differences between various versions of the llvm textual AST.
 data Config = Config { cfgLoadImplicitType :: Bool
                        -- ^ True when the type of the result of a load is
@@ -49,11 +47,11 @@ data Config = Config { cfgLoadImplicitType :: Bool
                      , cfgUseDILocation :: Bool
                      }
 
-withConfig :: Config -> (LLVM => a) -> a
+withConfig :: Config -> ((?config :: Config) => a) -> a
 withConfig cfg body = let ?config = cfg in body
 
 
-ppLLVM, ppLLVM35, ppLLVM36, ppLLVM37, ppLLVM38 :: (LLVM => a) -> a
+ppLLVM, ppLLVM35, ppLLVM36, ppLLVM37, ppLLVM38 :: ((?config :: Config) => a) -> a
 
 ppLLVM = ppLLVM38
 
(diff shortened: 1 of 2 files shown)


src/Text/LLVM/PP.hs
```diff
@@ -72,13 +70,26 @@ ppLLVM38 = withConfig Config { cfgLoadImplicitType = False
                              , cfgUseDILocation    = True
                              }
 
-checkConfig :: LLVM => (Config -> Bool) -> Bool
+checkConfig :: (?config :: Config) => (Config -> Bool) -> Bool
 checkConfig p = p ?config
 
 
+type Fmt a = (?config :: Config) => a -> Doc
```
RyanGlScott, line 81: My understanding is that the definition of this will change in a subsequent patch, but it would still be good to give it some rudimentary Haddocks for now.

src/Text/LLVM/PP.hs
```diff
@@ -72,13 +70,26 @@ ppLLVM38 = withConfig Config { cfgLoadImplicitType = False
...
+
+
+-- | The LLVMPretty class has instances for most AST elements.  It allows the
+-- conversion of an AST element (and its sub-elements) into a Doc assembly format
+-- by simply using the 'llvmPP' method rather than needing to explicitly invoke
+-- the specific pretty-printing function for that element.
+class LLVMPretty a where llvmPP :: Fmt a
+
+instance LLVMPretty Module where llvmPP = ppModule
+instance LLVMPretty Symbol where llvmPP = ppSymbol
```
RyanGlScott, line 91: There's not much payoff to adding this class just yet, since you don't call `llvmPP` anywhere. This isn't to say that I'm opposed to adding it here, but I think it remains to be seen if the payoff will be worth it in a subsequent patch.

RyanGlScott (approved):
I like it.

(For context: this is working towards a design where `Fmt` is used to query LLVM version–specific differences for pretty-printing. I reviewed an earlier version of this patch.)

src/Text/LLVM/PP.hs
```diff
@@ -74,6 +74,10 @@ checkConfig :: (?config :: Config) => (Config -> Bool) -> Bool
 checkConfig p = p ?config
 
 
+-- | This type encapsulates the ability to convert an object into Doc
+-- format. Using this abstraction allows for a consolidated representation of the
+-- declaration.  Most pretty-printing for LLVM elements will have a `Fmt a`
```
RyanGlScott, line 79: Haddock nitpick: backticks behave like single quotes in that they hyperlink particular identifiers. If you want just to render this in a monospace font, I think you want:

```hs
-- declaration.  Most pretty-printing for LLVM elements will have a @Fmt a@
```

Or, if you want `Fmt` to be hyperlinked:

```hs
-- declaration.  Most pretty-printing for LLVM elements will have a @'Fmt' a@
```

PR #119, opened 2023-08-04
Adds LLVMVer for version specification with ppLLVM and LLVMPretty Ident instance.

kquick:
This change adds an `LLVMVer` definition to allow specification of LLVM versions beyond the previous 3.5, 3.6, 3.7, and 3.8.  The `ppLLVM` function is updated to take this additional argument to determine the format of the LLVM output.

This also adds an `LLVMPretty` instance for `Ident`.

This is part of a series of pretty-printing updates, and more PR's will be forthcoming.

### DIFF
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -76,7 +76,7 @@ checkConfig p = p ?config
 
 -- | This type encapsulates the ability to convert an object into Doc
 -- format. Using this abstraction allows for a consolidated representation of the
--- declaration.  Most pretty-printing for LLVM elements will have a `Fmt a`
+-- declaration.  Most pretty-printing for LLVM elements will have a @'Fmt' a@
 -- function signature for that element.
 type Fmt a = (?config :: Config) => a -> Doc
 
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -34,6 +34,46 @@ import Prelude hiding ((<>))
 
 -- Pretty-printer Config -------------------------------------------------------
 
+
+-- | The value used to specify the LLVM major version.  The LLVM text format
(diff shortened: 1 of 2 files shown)


src/Text/LLVM/PP.hs
```diff
@@ -34,6 +34,46 @@ import Prelude hiding ((<>))
...
+llvmV3_5 = 0
+llvmV3_6 = 1
+llvmV3_7 = 2
+llvmV3_8 = 3
+
+-- | This value should be updated when support is added for new LLVM versions;
+-- this is used for defaulting and otherwise reporting the maximum LLVM version
+-- known to be supported.
+llvmVlatest :: LLVMVer
+llvmVlatest = 16
```
RyanGlScott, line 74: Is this used anywhere?

src/Text/LLVM/PP.hs
```diff
@@ -45,29 +85,43 @@ data Config = Config { cfgLoadImplicitType :: Bool
                        -- instruction is implied.
 
                      , cfgUseDILocation :: Bool
+
+                     , cfgVer :: LLVMVer
```
RyanGlScott, line 89: I'd be willing to bet that all of the fields above this one (`cfgLoadImplicitType`, `cfgGEPImplicitType`, and `cfgUseDILocation`) can be subsumed by checking `cfgVer`. I'm not sure if you were planning to do that in a separate PR, however.

RyanGlScott (approved):

src/Text/LLVM/PP.hs
kquick: It will be in forthcoming PR's (notably in llvm-pretty-bc-parser).

src/Text/LLVM/PP.hs
kquick: Yup, that's the plan.

PR #120, opened 2023-08-07
Specialize output on version only; add some tests for version-specific output.

kquick:


src/Text/LLVM/PP.hs
```diff
@@ -678,7 +648,7 @@ ppLoad ty ptr mo ma =
   where
   isAtomic = isJust mo
 
-  isImplicit = checkConfig cfgLoadImplicitType
+  isImplicit = llvmVer < llvmV3_7
```
RyanGlScott: This is a nit pick, but now that we are explicitly checking version numbers inline like this, I think I'd prefer phrasing this code like so:

```hs
ppLoad ty ptr mo ma =
  "load" <+> (if isAtomic   then "atomic" else empty)
         <+> (if isExplicit then explicit else empty)
         <+> ...
  where
  ...
  isExplicit = llvmVer >= llvmV3_7
```

This has a couple of benefits:

1. I prefer to check for the most recent version first so that if we ever have to refine this check further, it will read more straightforwardly:

   ```hs
     isExplicit
       | llvmVer >= 18
       = ...
       | llvmVer >= 17
       = ...
       | llvmVer >= llvmV3_7
       = ...
       ...
   ```
2. There is some added symmetry with the `isAtomic` check above.

src/Text/LLVM/PP.hs
```diff
@@ -785,7 +755,7 @@ ppGEP ib ty ptr ixs =
     <+> (if isImplicit then empty else explicit)
     <+> commas (map (ppTyped ppValue) (ptr:ixs))
   where
-  isImplicit = checkConfig cfgGEPImplicitType
+  isImplicit = llvmVer < llvmV3_7
```
RyanGlScott: Similarly, I'd prefer using `isExplicit = llvmVer >= llvmV3_7` here.

src/Text/LLVM/PP.hs
```diff
@@ -881,8 +851,8 @@ ppValMd :: Fmt ValMd
 ppValMd = ppValMd' ppLabel
 
 ppDebugLoc' :: Fmt i -> Fmt (DebugLoc' i)
-ppDebugLoc' pp dl = (if cfgUseDILocation ?config then "!DILocation"
-                                                 else "!MDLocation")
+ppDebugLoc' pp dl = (if llvmVer > llvmV3_6 then "!DILocation"
```
RyanGlScott: I'd prefer checking `llvmVer >= llvmV3_7` here so that you don't have to add 1 to the version when reading it.

RyanGlScott (approved):
Hooray!

RyanGlScott (approved):

PR #121, opened 2023-08-07
Fix issue #113 and #114

kquick:
Contains a fix for the quoting of the `DILabel` `name` field bug identified in Issue #114 .

Contains a fix for the `DICompileUnit` `rangesBaseAddress` (and subsequent) fields which was added in LLVM 11 as noted in Issue #113.  The fix does not switch the field type as proposed in the issue, but does update the pretty-printing.

### DIFF
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -1025,7 +1025,7 @@ ppDIBasicType bt = "!DIBasicType"
 
 ppDICompileUnit' :: Fmt i -> Fmt (DICompileUnit' i)
 ppDICompileUnit' pp cu = "!DICompileUnit"
-  <> parens (mcommas
+  <> parens (mcommas $
        [ pure ("language:"              <+> integral (dicuLanguage cu))
        ,     (("file:"                  <+>) . ppValMd' pp) <$> (dicuFile cu)
        ,     (("producer:"              <+>) . doubleQuotes . text)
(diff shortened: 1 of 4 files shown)


src/Text/LLVM/PP.hs
```diff
@@ -1046,12 +1043,15 @@ ppDICompileUnit' pp cu = "!DICompileUnit"
...
+       ]
+       ++ if llvmVer < 11 then [] else
+       [ pure ("rangesBaseAddress:"     <+> ppBool (dicuRangesBaseAddress cu))
        ,     (("sysroot:"               <+>) . doubleQuotes . text)
              <$> (dicuSysRoot cu)
        ,     (("sdk:"                   <+>) . doubleQuotes . text)
              <$> (dicuSDK cu)
-       ])
+       ]
+       )
```
RyanGlScott: I'd prefer `llvmVer >= 11`. Something like this, perhaps:

```hs
       [ pure ("rangesBaseAddress:"     <+> ppBool (dicuRangesBaseAddress cu))
       ,     (("sysroot:"               <+>) . doubleQuotes . text)
             <$> (dicuSysRoot cu)
       ,     (("sdk:"                   <+>) . doubleQuotes . text)
             <$> (dicuSDK cu)
       | llvmVer >= 11
       ]
```

src/Text/LLVM/PP.hs
```diff
@@ -1046,12 +1043,15 @@ ppDICompileUnit' pp cu = "!DICompileUnit"
        , pure ("splitDebugInlining:"    <+> ppBool (dicuSplitDebugInlining cu))
        , pure ("debugInfoForProfiling:" <+> ppBool (dicuDebugInfoForProf cu))
        , pure ("nameTableKind:"         <+> integral (dicuNameTableKind cu))
-       , pure ("rangesBaseAddress:"     <+> ppBool (dicuRangesBaseAddress cu))
+       ]
+       ++ if llvmVer < 11 then [] else
```
RyanGlScott: Perhaps worth adding a Haddock comment to `dicuRangesBaseAddress` ([here](https://github.com/elliottt/llvm-pretty/blob/e586747ce3f4a37245b56745b90d4c1c02423baf/src/Text/LLVM/AST.hs#L1498)) that it is only pretty-printed with LLVM 11 or later, the version that it was added.

src/Text/LLVM/PP.hs
```diff
@@ -971,7 +968,7 @@ ppDIImportedEntity = ppDIImportedEntity' ppLabel
 ppDILabel' :: Fmt i -> Fmt (DILabel' i)
 ppDILabel' pp ie = "!DILabel"
   <> parens (mcommas [ (("scope:"  <+>) . ppValMd' pp) <$> dilScope ie
-                     , pure ("name:" <+> text (dilName ie))
+                     , pure ("name:" <+> doubleQuotes (text (dilName ie)))
```
RyanGlScott: I think this would be preferable:

```suggestion
                     , pure ("name:" <+> ppStringLiteral (dilName ie))
```

As `ppStringLiteral` performs escaping of certain characters.

RyanGlScott (approved):
Hooray!

PR #127, opened 2023-09-23
Add missing DIDerivedType scope (didtScope) pretty printing field output

### LINKED ISSUE #126: Missing scope for DIDerivedType metadata in PP output.
The `didtScope` field was added to the AST, but the PP module was not updated.  This causes the `scope:` output to be omitted from `!DIDerivedType` output, which in turn causes the value to change from its previous value (e.g. `Just (ValMdRef 0)`) to `Nothing` if the AST for the former is pretty-printed and then that result is parsed by llvm-pretty-bc-parser.

kquick:
Fixes #126


RyanGlScott (approved):
Good catch!

PR #129, opened 2023-09-23
Do not emit external linkage or default visibility for global attrs.

### LINKED ISSUE #128: `llvm-as` fails on `@val = external default global i32 123`
Testing shows that the LLVM Project `llvm-as` assembler does not accept either `external` or `default` for these declarations.  It accepts: `@val = global i32 123`.

kquick:
Fixes #128

### DIFF
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -299,46 +299,21 @@ ppTypeDecl td = ppIdent (typeName td) <+> char '='
 
 ppGlobal :: Fmt Global
 ppGlobal g = ppSymbol (globalSym g) <+> char '='
-         <+> ppTheGlobalAttrs (globalAttrs g)
+         <+> ppGlobalAttrs (isJust $ globalValue g) (globalAttrs g)
          <+> ppType (globalType g) <+> ppMaybe ppValue (globalValue g)
           <> ppAlign (globalAlign g)
           <> ppAttachedMetadata (Map.toList (globalMetadata g))
-  where
-  isStruct | Just (ValStruct {}) <- globalValue g = True
-           | otherwise = False
-  ppTheGlobalAttrs | isStruct = ppStructGlobalAttrs
-                    | otherwise = ppGlobalAttrs
 
-ppGlobalAttrs :: Fmt GlobalAttrs
-ppGlobalAttrs ga
+ppGlobalAttrs :: Bool -> Fmt GlobalAttrs
+ppGlobalAttrs hasValue ga
     -- LLVM 3.8 does not emit or parse linkage information w/ hidden visibility
     | Just HiddenVisibility <- gaVisibility ga =
             ppVisibility HiddenVisibility <+> constant
+    | Just External <- gaLinkage ga
+    , Just DefaultVisibility <- gaVisibility ga
+    , hasValue = constant
     | otherwise =
-        -- In LLVM's llvm-as will not accept "default" or "external" for global
-        -- definitions of this type (for LLVM 5 through 16).  [There may be other
-        -- exceptions: this is all that has been empirically observed to date.]
-        case gaLinkage ga of
-          Just External -> id
-          l -> (ppMaybe ppLinkage l <+>)
-        $
-        case gaVisibility ga of
-          Just DefaultVisibility -> id
-          v -> (ppMaybe ppVisibility v <+>)
-        $
-        constant
-  where
-  constant | gaConstant ga = "constant"
-           | otherwise     = "global"
-
-ppStructGlobalAttrs :: Fmt GlobalAttrs
-ppStructGlobalAttrs ga
-    -- LLVM 3.8 does not emit or parse external linkage for
-    -- global structs
-    | Just External <- gaLinkage ga,
-      Just DefaultVisibility <- gaVisibility ga
-        = constant
-    | otherwise = ppGlobalAttrs ga
+        ppMaybe ppLinkage (gaLinkage ga) <+> ppMaybe ppVisibility (gaVisibility ga) <+> constant
   where
   constant | gaConstant ga = "constant"
            | otherwise     = "global"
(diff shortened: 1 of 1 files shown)


src/Text/LLVM/PP.hs
```diff
@@ -299,34 +299,21 @@ ppTypeDecl td = ppIdent (typeName td) <+> char '='
...
           <> ppAttachedMetadata (Map.toList (globalMetadata g))
-  where
-  isStruct | Just (ValStruct {}) <- globalValue g = True
-           | otherwise = False
-  ppTheGlobalAttrs | isStruct = ppStructGlobalAttrs
-                    | otherwise = ppGlobalAttrs
 
-ppGlobalAttrs :: Fmt GlobalAttrs
-ppGlobalAttrs ga
+ppGlobalAttrs :: Bool -> Fmt GlobalAttrs
```
RyanGlScott, line 313: Please document what this `Bool` means.

src/Text/LLVM/PP.hs
```diff
@@ -299,34 +299,21 @@ ppTypeDecl td = ppIdent (typeName td) <+> char '='
...
-ppStructGlobalAttrs ga
-    -- LLVM 3.8 does not emit or parse external linkage for
-    -- global structs
-    | Just External <- gaLinkage ga,
-      Just DefaultVisibility <- gaVisibility ga
-        = constant
-    | otherwise = ppGlobalAttrs ga
+    | Just External <- gaLinkage ga
+    , Just DefaultVisibility <- gaVisibility ga
+    , hasValue = constant
```
RyanGlScott: Please leave a comment explaining what this special case is checking for and why.

src/Text/LLVM/PP.hs
```diff
@@ -299,34 +299,21 @@ ppTypeDecl td = ppIdent (typeName td) <+> char '='
...
-ppStructGlobalAttrs :: Fmt GlobalAttrs
-ppStructGlobalAttrs ga
-    -- LLVM 3.8 does not emit or parse external linkage for
-    -- global structs
-    | Just External <- gaLinkage ga,
-      Just DefaultVisibility <- gaVisibility ga
-        = constant
-    | otherwise = ppGlobalAttrs ga
+    | Just External <- gaLinkage ga
+    , Just DefaultVisibility <- gaVisibility ga
```
RyanGlScott, line 319: If I understand this code correctly, this will only omit things if the global is both `external` _and_ `default`. What happens if it is `external` but not `default`, or `default` but not `external`—will `llvm-as` accept it then? (The phrasing "either `external` or `default`" on #128 led me to believe that that was not the case, but perhaps I misread that.)

src/Text/LLVM/PP.hs
kquick: done.

src/Text/LLVM/PP.hs
kquick: done.

src/Text/LLVM/PP.hs
kquick: Empirically, issue 128 is correct: neither is accepted, although issue #128 doesn't cover the case where there is no initial value.  The first commit I made on this PR handled 128 directly, but when I discovered that failed when it was truly an external value, I made adjustments, noticed that the result was nearly identical to the "structure" version, so I consolidated both here.  That may not be the correct resulting behavior, but so far I have never seen `external` without `default` (and vice-versa) based on empirical testing, and there is no other test framework to validate this functionality right now.

By empirically, I mean that I have tests (for llvm-pretty-bc-parser) that cover all 6 cases that I can think of:
* uniniitialized scalar
* initialized scalar
* uninitialized structure
* initialized structure
* external scalar
* external structure

These tests pass on my local copy of llvm-pretty-bc-parser/llvm-pretty that contain the aggregation of PRs 228 and 252 for the former and PRs 125, 127, 129, and 130 for the latter (all of which are needed to get things successful for LLVM 10 and 11).  I should have mentioned these and the fact that they are passing and my intent to commit them once this batch of PRs lands.

src/Text/LLVM/PP.hs
```diff
@@ -299,34 +299,26 @@ ppTypeDecl td = ppIdent (typeName td) <+> char '='
...
-           | otherwise = False
-  ppTheGlobalAttrs | isStruct = ppStructGlobalAttrs
-                    | otherwise = ppGlobalAttrs
 
-ppGlobalAttrs :: Fmt GlobalAttrs
-ppGlobalAttrs ga
+-- The first argument to ppGlobalAttrs indicates whether there is a value
+-- associated with this global declaration: a global declaration with a value
+-- should not be identified as "external" and "default" visibility, whereas one
+-- without a value may have those attributes.
```
RyanGlScott: You may as well make this a Haddock comment, since `ppGlobalAttrs` is exported from this module.

src/Text/LLVM/PP.hs
```diff
@@ -299,34 +299,26 @@ ppTypeDecl td = ppIdent (typeName td) <+> char '='
...
-  isStruct | Just (ValStruct {}) <- globalValue g = True
-           | otherwise = False
-  ppTheGlobalAttrs | isStruct = ppStructGlobalAttrs
-                    | otherwise = ppGlobalAttrs
 
-ppGlobalAttrs :: Fmt GlobalAttrs
-ppGlobalAttrs ga
+-- The first argument to ppGlobalAttrs indicates whether there is a value
+-- associated with this global declaration: a global declaration with a value
+-- should not be identified as "external" and "default" visibility, whereas one
```
RyanGlScott: To avoid Haddock escaping these as module identifiers:

```suggestion
-- should not be identified as \"external\" and \"default\" visibility, whereas one
```

src/Text/LLVM/PP.hs
RyanGlScott: OK, good to know. It would be worth leaving a comment next to this code explaining that the logic here is the result of empirical testing about places where `external`/`default` are permitted.

RyanGlScott (approved):

PR #130, opened 2023-09-24
DISubRange metadata extensions for recent LLVM versions.

kquick:


src/Text/LLVM/AST.hs
```diff
@@ -1647,10 +1647,14 @@ data DISubprogram' lab = DISubprogram
...
 
-data DISubrange = DISubrange
-  { disrCount      :: Int64
-  , disrLowerBound :: Int64
-  } deriving (Data, Eq, Generic, Ord, Show, Typeable)
+data DISubrange' lab = DISubrange
+  { disrCount      :: Either Int64 (Maybe (ValMd' lab))
+  , disrLowerBound :: Either Int64 (Maybe (ValMd' lab))
+  , disrUpperBound :: Maybe (ValMd' lab)
+  , disrStride     :: Maybe (ValMd' lab)
```
RyanGlScott, line 1677: If you happen to know this information, it would be helpful to document which LLVM version(s) introduced these new fields.

src/Text/LLVM/PP.hs
```diff
@@ -1235,11 +1235,20 @@ ppDISubprogram' pp sp = "!DISubprogram"
 ppDISubprogram :: Fmt DISubprogram
 ppDISubprogram = ppDISubprogram' ppLabel
 
+ppDISubrange' :: Fmt i -> Fmt (DISubrange' i)
+ppDISubrange' pp sr =
+  let eimb pfx l mbr = either (Just . (pfx <+>) . l) (fmap ((pfx <+>) . mbr))
```
RyanGlScott: It's really hard to grok what this code is doing at a glance, possibly because of all the pointfree code being smashed together. Perhaps adding an explicit type signature and/or a comment would help.

src/Text/LLVM/AST.hs
```diff
@@ -1647,10 +1647,14 @@ data DISubprogram' lab = DISubprogram
 
 type DISubprogram = DISubprogram' BlockLabel
 
-data DISubrange = DISubrange
-  { disrCount      :: Int64
-  , disrLowerBound :: Int64
-  } deriving (Data, Eq, Generic, Ord, Show, Typeable)
+data DISubrange' lab = DISubrange
+  { disrCount      :: Either Int64 (Maybe (ValMd' lab))
+  , disrLowerBound :: Either Int64 (Maybe (ValMd' lab))
```
RyanGlScott: Using `Either` here is somewhat grotesque. As an alternative, what about always encoding these as `Maybe (ValMd' lab)`? On old versions of LLVM, we can simply use `Just (ValMdValue (Typed (PrimType Integer) (ValInteger ...)))`

src/Text/LLVM/AST.hs
kquick: I don't at this point.  I'm basing this off the latest master code from here: https://github.com/llvm/llvm-project/blob/bbe8cd13335300958b04db5318c31ff52714f96f/llvm/lib/Bitcode/Reader/MetadataLoader.cpp#L1435-L1461. 

It *will* be important to know the versions for the pretty-printing aspect (which I discuss further below).  Right now I'd like to restore llvm-pretty-bc-parser/llvm-pretty to operational status (particularly vis-a-vis PR 228 in the former). Once we are operational for the current CI-verified llvm version range, I plan to widen the version range via CI to identify where these version-specific inflection points are for pretty-printing (and we now have the machinery in the CI to do this).

src/Text/LLVM/PP.hs
kquick: I simplified this overall and pushed an update.  Let me know if you think more commenting/help is needed with the simpler version.

src/Text/LLVM/AST.hs
kquick: Based on the comment from the llvm-project here: https://github.com/llvm/llvm-project/blob/bbe8cd13335300958b04db5318c31ff52714f96f/llvm/lib/Bitcode/Reader/MetadataLoader.cpp#L1437-L1444
it seems more authentic to the actual representation to use this Either encoding.  While we could potentially push the value down into the `ValMd` like you are suggesting, I think it obfuscates the dual-usage of this field and has possible issues for the pretty-printing side of things:

* It implies that any ValMd encoding could be used
* It creates partiality in the pretty-printing if a ValMd representation is chosen that cannot be represented as an inline signed integer (Integer -> ValMd is injective, but ValMd -> Integer is not).

As noted in my response to your previous comment above, the current pretty-printing will need to be enhanced to handle the (presumably older) LLVM versions that use the signed integer encoding rather than the ValMd... once we know what those versions are.  Right now there are about 6 PR's needing to land to get to that base operational point and I'd like to avoid introducing additional complexity until we get there.

src/Text/LLVM/AST.hs
RyanGlScott: I took a look at the `git blame` in LLVM:

* `discrCount` can be an arbitrary metadata node as of [LLVM 7](https://github.com/llvm/llvm-project/commit/fdf40917d946e3b36fd67afb2aac4063af9fc85d), which corresponds to `case 1` in the LLVM code.
* `discrLowerBound` can be an arbitrary metadata node as of [LLVM 11](https://github.com/llvm/llvm-project/commit/d20bf5a7258d4b6a7f017a81b125275dac1aa166), which corresponds to `case 2` in the LLVM code.
* `discrUpperBound` and `discrStride` were introduced in LLVM 11, in the same commit that changed `discrLowerBound`.

src/Text/LLVM/PP.hs
RyanGlScott: It's not quite as intimidating as before, but I still think adding an explicit type signature would go a long way here.

src/Text/LLVM/AST.hs
RyanGlScott: > * It implies that any ValMd encoding could be used

I don't think so? Even with the `Either Int64 (Maybe (ValMd' lab))` encoding, there are only a few possible `ValMd'` values that can appear here in practice. For example, the commit which [made `discrLowerBound` an arbitrary metadata node](https://github.com/llvm/llvm-project/commit/d20bf5a7258d4b6a7f017a81b125275dac1aa166) notes that it can only be a signed integer, a `DIVariable`, or a `DIExpression`. Either way, it is possible to have unrepresentable states. 

> * It creates partiality in the pretty-printing if a ValMd representation is chosen that cannot be represented as an inline signed integer (Integer -> ValMd is injective, but ValMd -> Integer is not).

I don't understand what you mean here: pretty-printing a `ValMd` is a total function. It's possible to have unrepresentable states that could be pretty-printed in ways that would result in invalid `discrLowerBound` values, but as noted above, that would happen regardless of how we encode these fields. I think it's better to just accept this, document what forms of `ValMd'` values are allowed here, and pick a simpler encoding (`Maybe (ValMd' lab)`).

src/Text/LLVM/AST.hs
```diff
@@ -1647,10 +1647,36 @@ data DISubprogram' lab = DISubprogram
...
-data DISubrange = DISubrange
-  { disrCount      :: Int64
-  , disrLowerBound :: Int64
-  } deriving (Data, Eq, Generic, Ord, Show, Typeable)
+-- | The DISubrange is a Value subrange specification, usually associated with
+-- arrays or enumerations.
+--
+-- * Early LLVM: only 'disrCount' and 'disrLowerBound' were present, where both
+--   were a direct signed 64-bit value.  This corresponds to "format 0" in the
+--   bitcode encoding.
```
RyanGlScott: I don't think "format 0" will mean anything to most readers unless you cite where this is used in the LLVM source code—worth doing so here.

src/Text/LLVM/AST.hs
```diff
@@ -1647,10 +1647,36 @@ data DISubprogram' lab = DISubprogram
...
+--
+-- * LLVM 11: 'disrLowerBound' was changed to a metadata representation and
+--   'disrUpperBound' and 'disrStride' were added (primarily driven by the
+--   addition of Fortran support in llvm).  All three should only be represented
+--   as a signed 64-bit integer, a Variable, or an Expression.  This corresponds
+--   to "format 2" in the bitcode encoding.  See
+--   https://github.com/llvm/llvm-project/commit/d20bf5a for this change.
+--
+-- Also see llvm-project/llvm/lib/Bitcode/Reader/MetadataLoader.cpp handling of
+-- bitc::METADATA_SUBRANGE (lines 1425-1461 in commit bbe8cd1) for how this is
```
RyanGlScott: To save users from having to dig up this information themselves, we should just give a permalink here:
https://github.com/llvm/llvm-project/blob/bbe8cd13335300958b04db5318c31ff52714f96f/llvm/lib/Bitcode/Reader/MetadataLoader.cpp#L1435-L1461

src/Text/LLVM/PP.hs
```diff
@@ -1235,11 +1235,34 @@ ppDISubprogram' pp sp = "!DISubprogram"
...
+             , let ppV = if llvmVer >= 11
+                         then ppValMd' pp
+                         else ppInt64ValMd'
+               in (("lowerBound:" <+>) . ppV) <$> disrLowerBound sr
+             , if llvmVer >= 11
+               then (("upperBound:" <+>) . ppValMd' pp) <$> disrUpperBound sr
+               else Nothing
+             , if llvmVer >= 11
+               then (("stride:" <+>) . ppValMd' pp) <$> disrStride sr
+               else Nothing
```
RyanGlScott: We could use [`when'`](https://github.com/elliottt/llvm-pretty/blob/3a9599072cb9fa59e9ba9321a9e46030ce975f43/src/Text/LLVM/PP.hs#L97-L100) here instead if these `if ... then ... else Nothing` constructs.

src/Text/LLVM/PP.hs
```diff
@@ -1279,6 +1302,30 @@ opt :: Bool -> Fmt Doc
...
+      Nothing -> mempty
+      Just v -> ppInt64ValMd' v
+  ValMdDebugInfo (DebugInfoGlobalVariableExpression expr) ->
+    case digveExpression expr of
+      Nothing -> mempty
+      Just e -> ppInt64ValMd' e
+  ValMdDebugInfo (DebugInfoLocalVariable lv) ->
+    integer $ fromIntegral $ dilvArg lv  -- ??
+  ValMdRef _idx -> mempty -- no table here to look this up...
+  _ -> mempty  -- TODO: generate warning?
```
RyanGlScott: This code is quite complicated, and (I think) needlessly so. Why not just `ppValMd'` instead of `ppInt64ValMd'`? This code seems to have aspirations of checking for invalid states, but I don't think the pretty-printing code is the right place to do this. (Besides, we don't check for invalid states in `count`/`lowerBound` on recent LLVM versions, so I don't really see the point in doing so here.)

src/Text/LLVM/AST.hs
kquick: Cited in the location of your comment below, but added a reference here.

src/Text/LLVM/AST.hs
kquick: Done.

src/Text/LLVM/PP.hs
kquick: Done.

src/Text/LLVM/PP.hs
kquick: This code is invoked for specific earlier versions of LLVM where the corresponding field is stored in the bitcode as a direct signed value instead of a metadata reference; using `ppValMd'` will emit a metadata reference.  This is not checking for invalid states, but instead this is the projection of a signed integer value back out of a `ValMd`.

If the PP is not the right location to adjust the output to be specific to the LLVM version we are emitting, where would you recommend doing this?

RyanGlScott: Thank you for pushing back here, since I now realize that the way I imagined that `DISubrange` should be pretty-printed was slightly off. In my mind, I had thought that pretty-printing as `Int64` value directly would result in the same `Doc` as pretty-printing a `ValMdValue (Typed (PrimType (Integer 64)) (ValInteger x))` value (which is what these `Int64` values get turned into on old LLVM versions). Now that I look more closely, however, that isn't the case at all. In `Text.LLVM.PP`, we have:

```hs
ppValMd' :: Fmt i -> Fmt (ValMd' i)
ppValMd' pp m = case m of
  ...
  ValMdValue tv     -> ppTyped (ppValue' pp) tv
  ...

ppTyped :: Fmt a -> Fmt (Typed a)
ppTyped fmt ty = ppType (typedType ty) <+> fmt (typedValue ty)

ppValue' :: Fmt i -> Fmt (Value' i)
ppValue' pp val = case val of
  ValInteger i       -> integer i
  ...
```

The key function here is `ppTyped`. This would cause, say, `ValMdValue (Typed (PrimType (Integer 64)) (ValInteger 27))` to be pretty-printed as `i64 27`, not `27`! This is definitely not what we want.

But looking even more closely, I think the same mistake also applies to the code is used on recent LLVM versions. Above in `ppDISubrange'`, we have:

```hs
                  let ppV = if llvmVer >= 7
                            then ppValMd' pp
                            else ppInt64ValMd'
                  in (("count:" <+>) . ppV) <$> disrCount sr
```

If I'm reading this correctly, then this code would have the potential to mistakenly print something like `count: i64 27` rather than `count: 27` on LLVM 7 and later. (I haven't tested this hypothesis, but it looks very likely.) Looking at the [LLVM repo's test cases](https://github.com/search?q=repo%3Allvm%2Fllvm-project+disubrange+language%3ALLVM&type=code&l=LLVM) (and the [code for pretty-printing `DISubrange`s](https://github.com/llvm/llvm-project/blob/431969ede13b0974041e948393d94bca1720c5d7/llvm/lib/IR/AsmWriter.cpp#L1888-L1927)), I don't think this was the intention.

As such, I've reluctantly come to the conclusion that we do need something like `ppInt64ValMd'`, and not just for old LLVM versions, but for _all_ LLVM versions. It's a bit grotesque, but then again, the [upstream LLVM code for pretty-printing `DISubrange`s](https://github.com/llvm/llvm-project/blob/431969ede13b0974041e948393d94bca1720c5d7/llvm/lib/IR/AsmWriter.cpp#L1888-L1927) is also grotesque, so it's hard to avoid.

kquick: IIUC, and based on the llvm-project's pretty printing (thanks for the reference), you are suggesting that ppInt64ValMd' should be updated such that the `mempty` case (or at least the last one on line 1326) should call `ppValMd'` instead; the net effect being we print a raw signed integer if possible, otherwise we fall back to a `!` ValMd output. [And possibly have a flag to disable that fallback for the lower LLVM versions?]

RyanGlScott: > you are suggesting that ppInt64ValMd' should be updated such that the `mempty` case (or at least the last one on line 1326) should call `ppValMd'` instead

Yes, I think something like that should suffice. I'll let you be the judge of how fancy the fallback logic should be for older LLVM versions.

kquick: OK, I adjusted for this, and it is passing all current tests (llvm-pretty and llvm-pretty-bc-parser).  We may need to revisit this as our tests and/or LLVM version ranges broaden, but I think this is a good starting point.  Thanks for your investigations on this as well: I think we are in a better spot for it.

src/Text/LLVM/PP.hs
```diff
@@ -1279,6 +1298,32 @@ opt :: Bool -> Fmt Doc
...
+              Nothing -> mempty
+              Just v -> go v
+          ValMdDebugInfo (DebugInfoGlobalVariableExpression expr) ->
+            case digveExpression expr of
+              Nothing -> mempty
+              Just e -> go e
+          ValMdDebugInfo (DebugInfoLocalVariable lv) ->
+            integer $ fromIntegral $ dilvArg lv  -- ??
+          -- ValMdRef _idx -> mempty -- no table here to look this up...
+          o -> if canFallBack then ppValMd' pp o else mempty
```
RyanGlScott: Possibly:

```suggestion
          o -> when' canFallBack $ ppValMd' pp o
```

src/Text/LLVM/PP.hs
```diff
@@ -1279,6 +1298,32 @@ opt :: Bool -> Fmt Doc
 opt True  = id
 opt False = const empty
 
+-- | Print a ValMd' value as a plain signed integer (Int64) if possible.  If the
+-- ValMd' is not representable as an Int64, defer to ValMd' printing (if
+-- canFallBack is True) or print nothing (for when a ValMd is not a valid
+-- representation).
+
+ppInt64ValMd' :: Bool -> Fmt i -> Fmt (ValMd' i)
+ppInt64ValMd' canFallBack pp = go
```
RyanGlScott, line 1307: I'm surprised that this function needs to be recursive at all. My general understanding of the code is that if there _is_ an integer literal, it will be represented as a top-level `ValMdValue ... ValInteger`, without being nested underneath some other metadata node. If that is indeed true, then all of the non-`ValMdValue ... ValInteger` cases ought to be able to just call out to `ppMdValue'`, right? (Sadly, I don't think any of our existing `llvm-pretty-bc-parser` test cases involve `DISubrange`s of this level of complexity.)

RyanGlScott (approved):
Thanks for being patient. I'm much happier with the new approach here. I've left two possible suggestions—feel free to incorporate them if you think they're reasonable.

src/Text/LLVM/PP.hs
kquick: Oh, good catch.

src/Text/LLVM/PP.hs
kquick: I think this is all verging into the gray area of things, including:
* Principled traversal to extract all possible Int64 representations v.s. typical implementation of the llvm-project C code
* Viability of expressing LLVM code from a newer version in an older syntax

If we consider an AST generated by llvm-pretty-bc-parser, it will always be the direct form, but the AST itself is exported and there's no guarantee a client might submit an AST that allowed an alternate format.  As pointed out previously on this PR, it's possible to express non-sensical things due to the broadness of the `ValMd` inherited from the llvm-project; an anathema from a typical Haskeller's perspective, but a practicality from the C-based llvm-project.

[I'm assuming you meant `ppValMd'` instead of `ppMdValue'`]

Since this is passing all current tests, I'm inclined to leave this for the moment to capture possible concerns.  As we broaden the tests in the future, hopefully we can add more `DISubrange` exercises as you mentioned.

PR #134, opened 2024-01-30
`llvm-pretty-0.12.0.0` release prep

RyanGlScott:
This is preparation for a Hackage release of `llvm-pretty-0.12.0.0`, which I will cut after this PR lands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,36 @@
+# `llvm-pretty`
+
+A pretty printing library that was inspired by the LLVM binding by Lennart
+Augustsson. The library provides a monadic interface to a pretty printer, that
+allows functions to be defined and called, generating the corresponding LLVM
+assembly when run.
+
+## LLVM language feature support
+
+Currently, `llvm-pretty` supports LLVM versions up through 17. As a result of
+the broad version coverage, the `llvm-pretty` AST is a superset of all versions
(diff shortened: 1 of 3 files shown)


langston-barrett (approved):

README.md
```diff
@@ -0,0 +1,29 @@
...
+assembly when run.
+
+## LLVM language feature support
+
+Currently, `llvm-pretty` supports LLVM versions up through 17. `llvm-pretty`
+strives to support a reasonable variety of [LLVM language
+features](https://llvm.org/docs/LangRef.html), but there are places where our
+coverage of the LLVM language is incomplete. If you need a LLVM feature that is
+not currently supported by `llvm-pretty`, please [file an
+issue](https://github.com/elliottt/llvm-pretty/issues/new).
```
kquick, line 22: Also worth noting here is that as a result of the broad version coverage, the AST is a superset of all llvm versions of the AST, so it is possible that the manner in which certain information is presented in the AST will be different depending on the llvm version originating the information, and conversely that it is possible to construct an AST that cannot be represented in a specific (or any) llvm version.

kquick (approved):

README.md
RyanGlScott: I've expanded this part of the `README` to include these caveats.

PR #136, opened 2024-02-26
fix printing of Infinites and NaNs

### LINKED ISSUE #135: Pretty-printing of infinite floating point and NaNs is incorrect/lossy
Specifically, this line is giving me trouble:

https://github.com/elliottt/llvm-pretty/blob/c4411152b05a59f0a4d6013c7f51178ab7368d81/src/Text/LLVM/PP.hs#L817

This uses [pretty's `double`](https://hackage.haskell.org/package/pretty-1.1.3.6/docs/src/Text.PrettyPrint.HughesPJ.html#double) which calls the [`show` instance for `Double`](https://hackage.haskell.org/package/base-4.19.1.0/docs/src/GHC.Float.html#formatRealFloatAlt), and for any `NaN` or `+/- Infinity`, it will print literally these strings.

This creates invalid LLVM IR, e.g.:
```
  %13 = fcmp oeq double %12, Infinity
```
which is not accepted by clang.

`llvm-dis` would instead create:
```
  %13 = fcmp oeq double %12, 0x7FF0000000000000
```
which recompiles correctly.

---

We could special case for `Nan`/`Infinity` the way the `show` instance does, and instead choose to grab the hexadecimal representation of the given choice of `Nan`/`Infinity` by doing:
```
text "0x" <> text (showHex (castDoubleToWord64 i) "")
```

But I almost wonder whether we just want to use this for all floats and doubles (even the representable ones).

Thoughts? @elliottt @RyanGlScott @kquick

Ptival:
fixes #135

In a nutshell, the old pretty-printing would output "Infinite" for infinite floating point values, which is not valid LLVM IR.  Instead, we must output the hexadecimal representation of the floating point value.

### DIFF
--- a/llvm-pretty.cabal
+++ b/llvm-pretty.cabal
@@ -46,7 +46,7 @@ Library
                        Text.LLVM.Triple.Parse.LookupTable
                        Text.LLVM.Util
 
-  Build-depends:       base             >= 4.9 && < 5,
+  Build-depends:       base             >= 4.11 && < 5,
                        containers       >= 0.4,
                        parsec           >= 3,
                        pretty           >= 1.0.1,
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -27,6 +27,7 @@ import Data.Char (isAlphaNum,isAscii,isDigit,isPrint,ord,toUpper)
 import Data.List (intersperse)
 import qualified Data.Map as Map
 import Data.Maybe (catMaybes,fromMaybe,isJust)
+import GHC.Float (castDoubleToWord64, castFloatToWord32)
 import Numeric (showHex)
 import Text.PrettyPrint.HughesPJ
 import Data.Int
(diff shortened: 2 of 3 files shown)


RyanGlScott:
Thanks! Would you be willing to add a couple of unit tests for this change?

Relatedly, [`castDoubleToWord64`](https://hackage.haskell.org/package/base-4.19.1.0/docs/GHC-Float.html#v:castDoubleToWord64) was only introduced in `base-4.11.0.0` (GHC 8.4), so we should bump the lower version bounds on `base` accordingly in `llvm-pretty.cabal`.

kquick:
There is no round-trip testing here, but adding some round-trip tests to `llvm-pretty-bc-parser` would be a nice bonus.

RyanGlScott:
To be clear, I wasn't imagining a full round-trip test, but rather something like `pretty (ValDouble 0x7FF0000000000000p0) == "0x7FF0000000000000"`.

Ptival:
How about [these tests](https://github.com/elliottt/llvm-pretty/pull/136/commits/19779497b91ff0c7cfa2fa72d0926ff0c5f826ea)?

RyanGlScott (approved):

PR #137, opened 2024-07-15
change pretty-printing of floats because of LLVM IR restrictions

Ptival:
As was pointed out by @tomsmeding here:
https://github.com/GaloisInc/llvm-pretty/issues/135#issuecomment-2227477305 we cannot use 32-bit (8 characters) "float" constants in LLVM IR.

Instead, we must embed the 32-bit float as 64-bit double constants with the same meaning.

### DIFF
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -27,7 +27,7 @@ import Data.Char (isAlphaNum,isAscii,isDigit,isPrint,ord,toUpper)
 import Data.List (intersperse)
 import qualified Data.Map as Map
 import Data.Maybe (catMaybes,fromMaybe,isJust)
-import GHC.Float (castDoubleToWord64, castFloatToWord32)
+import GHC.Float (castDoubleToWord64, float2Double)
 import Numeric (showHex)
 import Text.PrettyPrint.HughesPJ
 import Data.Int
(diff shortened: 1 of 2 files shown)


tomsmeding:
The particular case that triggered the problem for me is fixed with this PR, thanks! But this is only really a spot test, because this only accidentally popped up in the test suite of a larger codebase.

Thanks for the very quick PR!

src/Text/LLVM/PP.hs
```diff
@@ -816,9 +816,11 @@ ppValue' pp val = case val of
...
   -- Note: for +Inf/-Inf/NaNs, we want to output the bit-correct sequence
   ValFloat f         ->
-    if isInfinite f || isNaN f
-      then text "0x" <> text (showHex (castFloatToWord32 f) "")
-      else float f
+    -- WARNING: You should **not** use `castFloatToWord32` or `float` here.  LLVM IR does not
+    -- support 32-bit floating point constants, instead it wants to see 32-bit compatible 64-bit
+    -- constants.  We want to preserve the exact mantissa (zero-extended to the right from 23 to
+    -- 52 bits), which happens to be the behavior of `float2Double`.
+    text "0x" <> text (showHex (castDoubleToWord64 (float2Double f)) "")
```
RyanGlScott, line 823: After this PR, we will now always display `float`-typed constants in hexadecimal form, even values that aren't infinite or NaNs. Is that desirable? (As https://github.com/GaloisInc/llvm-pretty/issues/135#issuecomment-2227477305 shows, it's still possible to write things like `float 1.0`.)

Regardless of what choice we make here, we should add some test cases that exercise the pretty-printing code for non-infinite, non-NaN values at type `float` and `double`.

tomsmeding: Something that I apparently haven't run into yet, but as also mentioned in [the documentation](https://llvm.org/docs/LangRef.html#simple-constants), LLVM will only accept "readable" constants if they exactly match a floating-point value. In particular, `1.3` is accepted for `double` but not for `float`.
It may simply be easier to always output floating-point constants in hexadecimal form, so that we never run into issues like this. (And for my particular purpose, I don't care very much about readability of floating-point constants.)

RyanGlScott: Wow. That's quite tricky.

In that case, I agree that if there isn't a straightforward way to determine if LLVM will accept the output of calling the `float` function, then we should conservatively print all `float` constants in hexadecimal form. It would also be worth mentioning the exact floating-point value requirement in the comments here, as I wasn't aware of that additional subtlety until just now.

Ptival: Ah, if `float` prints things more nicely, and in a way that LLVM accepts, I can look into making this change less radical.

Ptival:
In a discussion with the group, we feel like we may just want to always print the hexadecimal version, possibly providing a flag to add a best-effort decimal expansion as a comment nearby where applicable. Setting this as a draft until I have more time to handle these considerations.

PR #141, opened 2024-08-28
Prepare release 0.13

mccleeary-galois:


kquick:
These changes seem to be bug fixes and enhancements: the existing API isn't changed in a non-backward-compatible way.  I'd recommend only a minor version bump (0.12.1) rather than a major version update, otherwise this will cause a significant amount of downstream cabal updates with no need for associated code changes.

Issue #144, opened 2024-08-31
Subtly incorrect logic in `Text.LLVM.call` ?


alpmestan:
Hello,

First off, thanks a lot for this package.

I stumbled upon some odd behaviour (possibly related to https://github.com/GaloisInc/llvm-pretty/issues/102#issuecomment-1478073111 ?). When I try to call a function that I either `declare`d or `define`d in the same `LLVM` monad computation, using the `Text.LLVM.call` function, the function type is printed with an excessive pointer, see below.

``` hs
mod1 = do
  printf <- L.declare (L.iT 32) "printf" [L.ptrT (L.iT 8)] True
  fmtStr <- L.string "d_fmt" "%lf\n\0"
  L.define L.emptyFunAttrs (L.iT 32) "main" () $ do
    _ <- L.call printf [fmtStr, doubleT L.-: L.toValue (pi :: Double)]
    L.ret (L.iT 32 L.-: (0 :: Int32))
```

This results in:

``` llvm
# test1.ll
target triple = "aarch64-apple-macosx-unknown-"
target datalayout = "e-m:o-i64:64-i128:128-n32:64-S128"
@d_fmt = constant [5 x i8] [i8 37, i8 108, i8 102, i8 10, i8 0]
declare i32 @printf(i8*, ...)
define i32 @main() {
  %r0 = call i32(i8*, ...)* @printf([5 x i8]* @d_fmt,
                                    double 3.141592653589793)
  ret i32 0
}
```

Notice the `*` at the end of `i32(i8*, ...)*`, which shouldn't be there.
This IR, when executed with `lli`:

``` sh
$ lli test1.ll
0.000000
```

If I edit the IR manually to remove that extra pointer in the type, and try to run again:

``` lllvm
# test2.ll
target triple = "aarch64-apple-macosx-unknown-"
target datalayout = "e-m:o-i64:64-i128:128-n32:64-S128"
@d_fmt = constant [5 x i8] [i8 37, i8 108, i8 102, i8 10, i8 0]
declare i32 @printf(i8*, ...)
define i32 @main() {
  %r0 = call i32 (i8*, ...) @printf([5 x i8]* @d_fmt,
                                    double 3.141592653589793)
  ret i32 0
}
```

``` sh
$ lli test2.ll
3.141593
```

I tracked down the issue to the definition of `call`:

``` hs
call :: IsValue a => Typed a -> [Typed Value] -> BB (Typed Value)
call sym vs = case typedType sym of
[...]

PR #145, opened 2024-09-03
make pretty-printing of call consistent with LLVM's expectations (#144)

alpmestan:
This addresses #144. Let me know if you'd like a test case for this.


kquick:
Thank you!

PR #146, opened 2024-09-30
Allow building with `th-abstraction-0.7.*`

RyanGlScott:
This makes it easier to obtain a build plan that is compatible with GHC 9.10.


sauclovian-g (approved):
Reviewing this kind of diff seems silly :-)

PR #148, opened 2024-11-26
Add support for LLVM address spaces

peterohanley:
Address spaces default to 0 and are currently mostly used by CHERI for capabilities (address space 200).

The thread_local property is properly read and reported.

Various formatting changes are implemented to more closely match llvm-dis output.

### DIFF
--- a/src/Text/LLVM.hs
+++ b/src/Text/LLVM.hs
@@ -414,7 +414,7 @@ iT :: Word32 -> Type
 iT  = PrimType . Integer
 
 ptrT :: Type -> Type
-ptrT  = PtrTo
+ptrT  = PtrTo (AddrSpace 0)
 
 voidT :: Type
 voidT  = PrimType Void
@@ -618,7 +618,7 @@ shuffleVector vec1 vec2 mask =
     _          -> error "shuffleVector not given a vector"
 
 alloca :: Type -> Maybe (Typed Value) -> Maybe Int -> BB (Typed Value)
-alloca ty mb align = observe (PtrTo ty) (Alloca ty es align)
+alloca ty mb align = observe (PtrTo (AddrSpace 0) ty) (Alloca ty (AddrSpace 0) es align)
   where
   es = fmap toValue `fmap` mb
 
@@ -628,7 +628,7 @@ load ty ptr ma = observe ty (Load ty (toValue `fmap` ptr) Nothing ma)
 store :: (IsValue a, IsValue b) => a -> Typed b -> Maybe Align -> BB ()
 store a ptr ma =
   case typedType ptr of
-    PtrTo ty -> effect (Store (ty -: a) (toValue `fmap` ptr) Nothing ma)
+    PtrTo adr ty -> effect (Store (ty -: a) (toValue `fmap` ptr) Nothing ma)
     _        -> error "store not given a pointer"
 
 nullPtr :: Type -> Typed Value
@@ -700,8 +700,8 @@ getelementptr ty ptr ixs = observe ty (GEP False ty (toValue `fmap` ptr) ixs)
 -- | Emit a call instruction, and generate a new variable for its result.
 call :: IsValue a => Typed a -> [Typed Value] -> BB (Typed Value)
 call sym vs = case typedType sym of
-  PtrTo ty@(FunTy rty _ _) -> observe rty (Call False ty (toValue sym) vs)
-  _                        -> error "invalid function type given to call"
+  PtrTo adr ty@(FunTy rty _ _) -> observe rty (Call False ty (toValue sym) vs)
+  _                            -> error "invalid function type given to call"
 
 -- | Emit a call instruction, but don't generate a new variable for its result.
 call_ :: IsValue a => Typed a -> [Typed Value] -> BB ()
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -93,6 +93,7 @@ module Text.LLVM.AST
     -- * Attributes
   , Linkage(..)
   , Visibility(..)
+  , ThreadLocality(..)
   , UnnamedAddr(..)
   , AddrSpace(..)
   , allocaAddrSpace
@@ -723,6 +724,7 @@ addGlobal g m = m { modGlobals = g : modGlobals m }
 data GlobalAttrs = GlobalAttrs
   { gaLinkage    :: Maybe Linkage
   , gaVisibility :: Maybe Visibility
+  , gaThreadLocality :: Maybe ThreadLocality
   , gaUnnamedAddr :: Maybe UnnamedAddr
   , gaAddrSpace  :: AddrSpace
   , gaConstant   :: Bool
@@ -871,6 +873,13 @@ data Linkage
 data Visibility = DefaultVisibility
                 | HiddenVisibility
                 | ProtectedVisibility
+    deriving (Data, Eq, Generic, Ord, Show, Typeable)
+
+data ThreadLocality = NotThreadLocal
+                | ThreadLocal
+                | LocalDynamic
+                | InitialExec
+                | LocalExec
     deriving (Data, Eq, Generic, Ord, Show, Typeable)
 
 -- ^ there is also None, use Nothing for it
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -394,7 +394,18 @@ data SelectionKind = ComdatAny
 -- Identifiers -----------------------------------------------------------------
 
 newtype Ident = Ident String
-    deriving (Data, Eq, Generic, Ord, Show, Typeable, Lift)
+    deriving (Data, Eq, Generic, Show, Typeable, Lift)
+
+-- this exists so that struct types will be printed in the same order llvm-dis prints them
+instance Ord Ident where
+  compare (Ident l) (Ident r) = go l r where
+    go [] [] = EQ
+    go [] _ = GT
+    go _ [] = LT
+    go (ll:ls) (rr:rs) = case compare ll rr of
+      LT -> LT
+      GT -> GT
+      EQ -> go ls rs
 
 instance IsString Ident where
   fromString = Ident
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -93,6 +93,11 @@ module Text.LLVM.AST
     -- * Attributes
   , Linkage(..)
   , Visibility(..)
+  , UnnamedAddr(..)
+  , AddrSpace(..)
+  , allocaAddrSpace
+  , programAddrSpace
+  , ptrAddrSpace
   , GC(..)
     -- * Typed Things
   , Typed(..)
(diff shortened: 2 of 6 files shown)


src/Text/LLVM.hs
```diff
@@ -414,7 +414,7 @@ iT :: Word32 -> Type
 iT  = PrimType . Integer
 
 ptrT :: Type -> Type
-ptrT  = PtrTo
+ptrT  = PtrTo (AddrSpace 0)
```
peterohanley, line 417: Hardcoded 0 address space here and elsewhere is reasonable because every previous use was with 0 (nothing else existed) but additional helpers may be useful later.

src/Text/LLVM/AST.hs
```diff
@@ -841,6 +875,22 @@ data Visibility = DefaultVisibility
...
+    deriving (Data, Eq, Generic, Ord, Show, Typeable)
+
+-- ^ there is also None, use Nothing for it
+data UnnamedAddr = GlobalUnnamedAddr
+                 | LocalUnnamedAddr
+    deriving (Data, Eq, Generic, Ord, Show, Typeable)
+
+newtype AddrSpace = AddrSpace
+  { getAddrSpace :: Int
+  } deriving (Data, Eq, Generic, Ord, Show, Typeable) -- , Bits, FiniteBits)
```
RyanGlScott, line 892: Would you be willing to derive a `Num` instance for this? It would be convenient to be able to write, say, `0` instead of `AddrSpace 0`.

RyanGlScott, line 892: Can the commented-out code here be removed, or is there a need to keep it around?

src/Text/LLVM/AST.hs
```diff
@@ -841,6 +875,22 @@ data Visibility = DefaultVisibility
...
     deriving (Data, Eq, Generic, Ord, Show, Typeable)
 
+data ThreadLocality = NotThreadLocal
+                | ThreadLocal
+                | LocalDynamic
+                | InitialExec
+                | LocalExec
+    deriving (Data, Eq, Generic, Ord, Show, Typeable)
+
+-- ^ there is also None, use Nothing for it
```
RyanGlScott, line 885: What is this comment referring to?

src/Text/LLVM.hs
RyanGlScott: Could you factor `AddrSpace 0` out into its own definition? Something like this (with an appropriate Haddock comment):

```hs
defaultAddrSpace :: AddrSpace
defaultAddrSpace = Addr 0
```

And then use this in places where `AddrSpace 0` is currently used. This would make it more obvious at a glance that `AddrSpace 0` is a distinguished value, IMO.

src/Text/LLVM/AST.hs
```diff
@@ -284,7 +290,7 @@ type DataLayout = [LayoutSpec]
 data LayoutSpec
   = BigEndian
   | LittleEndian
-  | PointerSize   !Int !Int !Int (Maybe Int) -- ^ address space, size, abi, pref
+  | PointerSize   !Bool !Int !Int !Int (Maybe Int) (Maybe Int) -- ^ fat pointer?, address space, size, abi, pref, "size of index used in GEP for address calculation"
```
RyanGlScott, line 293: Interestingly, the first `Int` here is intended to be an address space, but it is not using the `AddrSpace` type. I wonder if it should.

RyanGlScott, line 293: Where is the `Bool` (fat pointer?) documented? I'm looking [here](https://llvm.org/docs/LangRef.html#data-layout) in the LLVM Language Reference Manual, but I don't see anything about fat pointers in the `p[n]:<size>:<abi>[:<pref>][:<idx>]` documentation.

RyanGlScott, line 293: > "size of index used in GEP for address calculation"

I'm not sure what to make of the quotation marks here. Is this quoting documentation from somewhere? If so, we should escape these double quotes, since Haddock will try to render things between double quotes as Haskell module names, which likely isn't what you intended here.

src/Text/LLVM/AST.hs
```diff
@@ -293,6 +299,9 @@ data LayoutSpec
   | NativeIntSize [Int]
   | StackAlign    !Int -- ^ size
   | Mangling Mangling
+  | AllocaAddressSpace !AddrSpace
+  | DefaultGlobalVariableAddressSpace !AddrSpace
```
RyanGlScott, line 303: Perhaps we should abbreviate this to `GlobalVariableAddressSpace`? It might just be my reading of it, but I think the "`Default`" part is implied, and it's already a pretty long name.

src/Text/LLVM/AST.hs
```diff
@@ -380,7 +395,18 @@ data SelectionKind = ComdatAny
...
+-- this exists so that struct types will be printed in the same order llvm-dis prints them
+instance Ord Ident where
+  compare (Ident l) (Ident r) = go l r where
+    go [] [] = EQ
+    go [] _ = GT
+    go _ [] = LT
+    go (ll:ls) (rr:rs) = case compare ll rr of
+      LT -> LT
+      GT -> GT
+      EQ -> go ls rs
```
RyanGlScott, line 409: Out of curiosity, how did you figure out that this is the approach that `llvm-dis` used? If there is LLVM documentation (or even code) that we can cite as inspiration for this, that would be ideal.

src/Text/LLVM/AST.hs
```diff
@@ -841,6 +875,22 @@ data Visibility = DefaultVisibility
                 | ProtectedVisibility
     deriving (Data, Eq, Generic, Ord, Show, Typeable)
 
+data ThreadLocality = NotThreadLocal
+                | ThreadLocal
+                | LocalDynamic
+                | InitialExec
+                | LocalExec
```
RyanGlScott, line 882: Style nit pick: I'd prefer if `NotThreadLocal` was aligned with the other data constructors. Perhaps something like:

```suggestion
data ThreadLocality
  = NotThreadLocal
  | ThreadLocal
  | LocalDynamic
  | InitialExec
  | LocalExec
```

src/Text/LLVM/AST.hs
```diff
@@ -1822,3 +1873,21 @@ resolveValueIndex ty is@(ix:ixs) = case ty of
...
+programAddrSpace (ProgramAddressSpace as:_) = as
+programAddrSpace (_:ds) = programAddrSpace ds
+
+ptrAddrSpace :: Typed a -> AddrSpace
+ptrAddrSpace = go . typedType
+  where
+    go :: Type -> AddrSpace
+    go (PtrTo x _) = x
+    go (PtrOpaque x) = x
+    go _ = AddrSpace 0
```
RyanGlScott, line 1893: It would be worth leaving a Haddock comment saying that this returns the default address space on all non-pointer types (and this includes alias types).

src/Text/LLVM/PP.hs
```diff
@@ -227,6 +230,18 @@ ppLayoutBody size abi mb = int size <> char ':' <> int abi <> pref
...
+-- | Pretty-print the common case for data layout specifications.
+ppLayoutBody' :: Int -> Int -> Maybe Int -> Fmt (Maybe Int)
+ppLayoutBody' size abi mb mg = int size <> char ':' <> int abi <> pref <> gep
+  where
+  pref = case mb of
+    Nothing -> empty
+    Just p  -> char ':' <> int p
+  gep = case (mb, mg) of
+    (Just _, Just g) -> char ':' <> int g
+    _ -> empty
```
RyanGlScott, line 242: We now have two functions for pretty-printing data layout bodies (`ppLayoutBody` and `ppLayoutBody'`), both of which claim to be "the common case" in their Haddocks. This seems somewhat suspect. Instead of having two functions, could we simply add a `Maybe Int` argument to `ppLayoutBody`?

src/Text/LLVM/PP.hs
```diff
@@ -338,9 +353,15 @@ ppGlobalAttrs hasValue ga
         --   * initialized structure
         --   * external scalar
         --   * external structure
+        ppAddrSpace (gaAddrSpace ga) <+>
         constant
```
RyanGlScott, line 357: The comments above this ought to be updated, since it claims that only the value (and nothing else) is printed. Really, it ought to say that the address space and the value are printed now.

src/Text/LLVM/PP.hs
```diff
@@ -378,11 +400,12 @@ ppDefine d = "define"
          <+> ppSymbol (defName d)
           <> ppArgList (defVarArgs d) (map (ppTyped ppIdent) (defArgs d))
          <+> hsep (ppFunAttr <$> defAttrs d)
+         <+> ppAddrSpace (defAddrSpace d)
          <+> ppMaybe (\s  -> "section" <+> doubleQuotes (text s)) (defSection d)
          <+> ppMaybe (\gc -> "gc" <+> ppGC gc) (defGC d)
          <+> ppMds (defMetadata d)
          <+> char '{'
-         $+$ vcat (map ppBasicBlock (defBody d))
+         $+$ vcat (punctuate "" $ map ppBasicBlock (defBody d))
```
RyanGlScott, line 408: What is the motivation for this change?

src/Text/LLVM/PP.hs
```diff
@@ -478,13 +501,28 @@ ppLinkage linkage = case linkage of
 
 ppVisibility :: Fmt Visibility
 ppVisibility v = case v of
-    DefaultVisibility   -> "default"
+    DefaultVisibility   -> empty
```
RyanGlScott, line 504: What is the motivation for this change?

src/Text/LLVM/PP.hs
```diff
@@ -947,7 +984,7 @@ ppConstExpr' pp expr =
     ConstArith      op a b -> ppArithOp op <+> ppTuple a b
     ConstUnaryArith op a   -> ppUnaryArithOp op <+> ppTyp' a
     ConstBit        op a b -> ppBitOp op   <+> ppTuple a b
-  where ppTuple  a b = parens $ ppTyped ppVal' a <> comma <+> ppVal' b
+  where ppTuple  a b = parens $ ppTyped ppVal' a <> comma <+> ppTyped ppVal' (const b <$> a)
```
RyanGlScott, line 987: What is the motivation for this change?

src/Text/LLVM/AST.hs
```diff
@@ -841,6 +875,22 @@ data Visibility = DefaultVisibility
...
+                | InitialExec
+                | LocalExec
+    deriving (Data, Eq, Generic, Ord, Show, Typeable)
+
+-- ^ there is also None, use Nothing for it
+data UnnamedAddr = GlobalUnnamedAddr
+                 | LocalUnnamedAddr
+    deriving (Data, Eq, Generic, Ord, Show, Typeable)
+
+newtype AddrSpace = AddrSpace
```
RyanGlScott, line 890: The comments in [this part](https://llvm.org/docs/LangRef.html#data-layout) of the LLVM Language Reference Manual suggest that an address space's value must be less than `2^24`. Is that worth putting in the Haddocks here?

RyanGlScott (commented):
I've done an initial pass (without having yet looked at the corresponding `llvm-pretty-bc-parser` changes).

src/Text/LLVM/AST.hs
kquick: It might also be good to point out how this is different from the default `Ord`; I think it's that the `GT`/`LT` for an empty `Ident` array is inverted from the norm?

src/Text/LLVM/AST.hs
```diff
@@ -698,13 +724,18 @@ addGlobal g m = m { modGlobals = g : modGlobals m }
...
+  , gaUnnamedAddr :: Maybe UnnamedAddr
+  , gaAddrSpace  :: AddrSpace
   , gaConstant   :: Bool
   } deriving (Data, Eq, Generic, Ord, Show, Typeable)
 
 emptyGlobalAttrs :: GlobalAttrs
 emptyGlobalAttrs  = GlobalAttrs
   { gaLinkage    = Nothing
   , gaVisibility = Nothing
+  , gaUnnamedAddr = Nothing
```
kquick, line 737: Missing `gaThreadLocality = Nothing`?

PR #149, opened 2024-12-08
Add more haddocks for AST Instructions

kquick:


src/Text/LLVM/AST.hs
```diff
@@ -1196,7 +1196,17 @@ data Instr' lab
 
 
   | ShuffleVector (Typed (Value' lab)) (Value' lab) (Typed (Value' lab))
-
+    {- ^ * Constructs a fixed permutation of two input vectors: the first
+           argument is the input vector V1, the second input vector V2, and
+           the third is the mask,
+           the third is the insert position.
```
RyanGlScott: This part of the sentence is worded confusingly, as it describes "the third" twice in a row, but with different descriptions. Perhaps just shorten this to "the third argument is the mask".

src/Text/LLVM/AST.hs
```diff
@@ -1196,7 +1196,17 @@ data Instr' lab
 
 
   | ShuffleVector (Typed (Value' lab)) (Value' lab) (Typed (Value' lab))
-
+    {- ^ * Constructs a fixed permutation of two input vectors: the first
+           argument is the input vector V1, the second input vector V2, and
```
RyanGlScott: This goes through the effort of naming the input vectors V1 and V2, but it then never uses the names V1 or V2 in the rest of the Haddocks. Perhaps shorten this to "the first and second arguments are the input vectors, and the third argument is the mask".

src/Text/LLVM/AST.hs
```diff
@@ -1208,6 +1218,16 @@ data Instr' lab
...
   | Invoke Type (Value' lab) [Typed (Value' lab)] lab lab
+    {- ^ * Calls the specified target function, then branches to the success
+           label.  If an exception occurs during the call, the exception unwind
+           handling branches to the second label.
+         * Arguments:
+           1. The FunctionType
+           2. The Function target
+           3. arguments
+           4. onSuccess ret target label
+           5. onException unwind target label
```
RyanGlScott: There is some rather unusual capitalization going on here, between:

* "FunctionType" (is that supposed to be code?)
* "The Function target" (why is Function randomly capitalized?)
* "onSuccess"/"onException" (why do these randomly use lowerCamelCase?)

I thought that perhaps these might have been taken from the [LLVM Language Reference Manual entry](https://releases.llvm.org/17.0.1/docs/LangRef.html#invoke-instruction) for `invoke`, but I don't see anything of the sort there, so I'm not sure where these conventions came from.

src/Text/LLVM/AST.hs
```diff
@@ -1217,21 +1237,49 @@ data Instr' lab
 
   | Unwind
   | VaArg (Typed (Value' lab)) Type
+    -- ^ Accesses arguments passed through "varargs" areas of a function call.
+    -- The argument is a va_list*; this instruction returns the value of the
+    -- specified type located at the target and increments the pointer.
```
RyanGlScott: Minor formatting nit picks:

```suggestion
    -- ^ Accesses arguments passed through \"varargs\" areas of a function call.
    -- The argument is a @va_list*@; this instruction returns the value of the
    -- specified type located at the target and increments the pointer.
```

src/Text/LLVM/AST.hs
```diff
@@ -1217,21 +1237,49 @@ data Instr' lab
...
   | Switch (Typed (Value' lab)) lab [(Integer,lab)]
-    {- ^ * Multi-way branch: the first value determines the direction
-           of the branch, the label is a default direction, if the value
-           does not appear in the jump table, the last argument is the
-           jump table.
+    {- ^ * Multi-way branch: the first value determines the target index
+           for the jump, which is looked up in the third argument table
+           (key values are unique).  If the value is not found in the table,
+           the second argument is the default destination if the target is
+           not in the table.
```
RyanGlScott: Part of this sentence is redundant. Perhaps:

```suggestion
           (key values are unique). The second argument is the default
           destination if the target is not in the table.
```

src/Text/LLVM/AST.hs
```diff
@@ -1217,21 +1237,49 @@ data Instr' lab
...
+           1. The result type (the values set by the personality function
+              on re-entry to the function).
+           2. the second argument may be the personality function, which defines
+              values on re-entry. This is used in older LLVM versions and is
+              not supplied for recent LLVM versions.
+           3. True if this block is a "cleanup"
+           4. The list of clauses to handle the exception;  The clauses are
+              used to match the exception thrown.
+         * If no clause matches and cleanup not set, continue unwinding up
+           the stack (see Resume).
```
RyanGlScott: Minor formatting nit picks:

```suggestion
    {- ^ Target of an exception (from the 'Invoke' instruction).
         * Arguments:
           1. The result type (the values set by the personality function
              on re-entry to the function).
           2. the second argument may be the personality function, which defines
              values on re-entry. This is used in older LLVM versions and is
              not supplied for recent LLVM versions.
           3. True if this block is a "cleanup"
           4. The list of clauses to handle the exception;  The clauses are
              used to match the exception thrown.
         * If no clause matches and cleanup not set, continue unwinding up
           the stack (see 'Resume').
```

src/Text/LLVM/AST.hs
```diff
@@ -1217,21 +1237,49 @@ data Instr' lab
...
+         * If cleanup is false, there must be at least one clause
+     -}
 
   | Resume (Typed (Value' lab))
+    {- ^ Resumes propagation of an in-flight exception whose unwinding was
+         interrupted by a LandingPad instruction.
```
RyanGlScott: Minor formatting nit pick:

```suggestion
         interrupted by a 'LandingPad' instruction.
```

src/Text/LLVM/AST.hs
```diff
@@ -1217,21 +1237,49 @@ data Instr' lab
...
+         * Argument: the value of the exception to propagate.
+    -}
 
   | Freeze (Typed (Value' lab))
     {- ^ * Used to stop propagation of @undef@ and @poison@ values.
+         * If the argument is undef or poison, returns an arbitrary (but fixed)
+           value of that type instead, otherwise a no-op and returns its
+           argument.
```
RyanGlScott: For consistency:

```suggestion
    {- ^ * Used to stop propagation of @undef@ and @poison@ values.
         * If the argument is @undef@ or @poison@, returns an arbitrary (but fixed)
           value of that type instead, otherwise a no-op and returns its
           argument.
```

src/Text/LLVM/AST.hs
kquick: Oops, yes, the second line is leftover garbage.

kquick:
Sorry about the bad syntax: I was making these changes as a side-effect of some other work and decided it would be better to commit them than discard them, but I did the former without proofreading and converting in-progress to final product. Hopefully these look better now.

RyanGlScott (approved):

PR #150, opened 2024-12-08
More parameterization for functions not needing a specialized type parameter.

kquick:
Weakens type declarations to be more permissive where the stricter type isn't needed.  This  should not affect any existing uses, but permits a broader utilization of these functions where desireable.

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -480,11 +480,11 @@ isFloatingPoint :: PrimType -> Bool
 isFloatingPoint (FloatType _) = True
 isFloatingPoint _             = False
 
-isAlias :: Type -> Bool
+isAlias :: Type' ident -> Bool
 isAlias Alias{} = True
 isAlias _       = False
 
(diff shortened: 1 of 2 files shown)


src/Text/LLVM/AST.hs
```diff
@@ -480,11 +480,11 @@ isFloatingPoint :: PrimType -> Bool
 isFloatingPoint (FloatType _) = True
 isFloatingPoint _             = False
 
-isAlias :: Type -> Bool
+isAlias :: Type' ident -> Bool
 isAlias Alias{} = True
 isAlias _       = False
 
-isPrimTypeOf :: (PrimType -> Bool) -> Type -> Bool
+isPrimTypeOf :: (PrimType -> Bool) -> Type' lab -> Bool
```
RyanGlScott: The code here is inconsistent about calling `Type`'s type parameter `ident` or `lab` (and similarly with the code below). I think we should pick one convention and stick to it.

RyanGlScott (approved):
This seems reasonable, but do make sure to advertise these changes in the changelog.

PR #151, opened 2025-03-05
ci: Update `cache` action from v1 (deprecated) to v4

langston-barrett:


.github/workflows/haskell.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         cabal-version: ${{ matrix.cabal }}
 
     - name: Cache
-      uses: actions/cache@v1
+      uses: actions/cache@v4
```
RyanGlScott, line 28: `actions/checkout` (used above) could also stand to be updated to `@v4`.

RyanGlScott (approved):

PR #154, opened 2025-03-20
Prepare release 0.13.0

sauclovian-g:


RyanGlScott (approved):

langston-barrett:
While we're here, could you update the `tested-with` field of the Cabal file to match CI?

https://github.com/GaloisInc/llvm-pretty/blob/b4c252c831c44b70d36a715e6d7bcee5676ddef7/llvm-pretty.cabal#L16

https://github.com/GaloisInc/llvm-pretty/blob/b4c252c831c44b70d36a715e6d7bcee5676ddef7/.github/workflows/haskell.yml#L16

sauclovian-g:
Good catch! (and it's not just out of date, it's way out of date)

PR #163, opened 2025-07-19
Support function pointer alignment in data layouts

RyanGlScott:
This performs the `llvm-pretty`-side changed needed towards a resolution for https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/292.

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -28,6 +28,7 @@ module Text.LLVM.AST
     -- * Data Layout
   , DataLayout
   , LayoutSpec(..)
+  , FunctionPointerAlignType(..)
   , Mangling(..)
   , parseDataLayout
     -- * Inline Assembly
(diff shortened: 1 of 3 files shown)


src/Text/LLVM/AST.hs
```diff
@@ -342,6 +355,14 @@ parseDataLayout str =
            'm' -> Mangling      <$> (char ':' >> pMangling)
            _   -> mzero
 
+    pFunctionPointerAlignType :: Parser FunctionPointerAlignType
+    pFunctionPointerAlignType =
+      do c <- letter
+         case c of
+           'i' -> return IndependentOfFunctionAlign
+           'n' -> return IndependentOfFunctionAlign
```
kquick: Should this be `'n' -> return MultipleOfFunctionAlign` ?

kquick (dismissed):
Not sure about that case statement, esp. relative to the one in the  pretty-printer function.

kquick:
Also, the failure in llvm-pretty-bc-parser indicates you might need a `Generic` instance here.

langston-barrett (commented):
LGTM once the inconsistency is addressed!

src/Text/LLVM/AST.hs
RyanGlScott: Oops, good catch! That is a copy-paste mistake, and it should indeed be `MultipleOfFunctionAlign`.

RyanGlScott:
> Also, the failure in llvm-pretty-bc-parser indicates you might need a `Generic` instance here.

I assume you mean an `Arbitrary` instance, as indicated in [this failing CI job](https://github.com/GaloisInc/llvm-pretty-bc-parser/actions/runs/16393326965/job/46322177913?pr=298)? Currently, we define orphan `Arbitrary` instances on the `llvm-pretty-bc-parser` side (see [this code](https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/4297edff2501654b61fab997fc58d1394e3af309/unit-test/Tests/Instances.hs#L31-L32)), so I'll make that change in https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/298.

langston-barrett (approved):

PR #165, opened 2025-08-11
ci: Support GHC 9.12.3, drop 9.6.7

### LINKED ISSUE #156: Support GHC 9.12
https://www.haskell.org/ghc/download_ghc_9_12_1.html

Steps:

- [ ] Wait for a point-release that includes https://gitlab.haskell.org/ghc/ghc/-/issues/25771 (9.12.3)
- [ ] Allow the new version of `base` in the Cabal `build-depends`
- [ ] Run `cabal {build,test,haddock}`, bumping dependency bounds as needed
- [ ] Fix any new warnings from `-Wdefault`
- [ ] Add the new GHC version to the matrix in the GitHub Actions configuration
- [ ] Add the new GHC version to the Cabal `tested-with` field
- [ ] Optionally follow the below steps to remove any old GHC versions
  - [ ] Remove the old version from the matrix in the GitHub Actions configuration
  - [ ] Remove the old version from the Cabal `tested-with` field
  - [ ] Remove outdated CPP `ifdef`s that refer to the dropped version
  - [ ] Remove outdated `if` stanzas in the Cabal file

langston-barrett:
Fixes #156.

### DIFF
--- a/src/Text/LLVM/Triple/AST.hs
+++ b/src/Text/LLVM/Triple/AST.hs
@@ -21,7 +21,6 @@ module Text.LLVM.Triple.AST
   ) where
 
 import Data.Data (Data)
-import Data.Typeable (Typeable)
 import GHC.Generics (Generic)
 
 -- | The constructors of this type exactly mirror the LLVM @enum ArchType@,
@@ -150,7 +149,7 @@ data Arch
   | RenderScript64
     -- | NEC SX-Aurora Vector Engine
   | VE
-  deriving (Bounded, Data, Eq, Enum, Generic, Ord, Read, Show, Typeable)
+  deriving (Bounded, Data, Eq, Enum, Generic, Ord, Read, Show)
 
 -- | A 'First'-like semigroup instance that simply drops the RHS, unless the LHS
 -- is 'UnknownArch'.
@@ -216,7 +215,7 @@ data SubArch
   | SPIRVSubArch_v13
   | SPIRVSubArch_v14
   | SPIRVSubArch_v15
-  deriving (Bounded, Data, Enum, Eq, Generic, Ord, Read, Show, Typeable)
+  deriving (Bounded, Data, Enum, Eq, Generic, Ord, Read, Show)
 
 -- | A 'First'-like semigroup instance that simply drops the RHS, unless the LHS
 -- is 'NoSubArch'.
@@ -250,7 +249,7 @@ data Vendor
   | Mesa
   | SUSE
   | OpenEmbedded
-  deriving (Bounded, Data, Enum, Eq, Generic, Ord, Read, Show, Typeable)
+  deriving (Bounded, Data, Enum, Eq, Generic, Ord, Read, Show)
 
 -- | A 'First'-like semigroup instance that simply drops the RHS, unless the LHS
 -- is 'UnknownVendor'.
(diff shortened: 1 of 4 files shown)


langston-barrett:
Whoops, forgot that we're waiting on 9.12.3 [(which is scheduled for August)](https://gitlab.haskell.org/ghc/ghc/-/milestones/407#tab-issues) due to  https://gitlab.haskell.org/ghc/ghc/-/issues/25771.

langston-barrett:
Apparently, GHC 9.12.3 is now expected "the week of 24 November 2025".

https://www.haskell.org/ghc/blog/20251111-ghc-9.12.3-rc2-released.html

langston-barrett:
GHC 9.12.3 [is now available](https://www.haskell.org/ghc/blog/20251227-ghc-9.12.3-released.html).

src/Text/LLVM/Triple/AST.hs
```diff
@@ -21,7 +21,6 @@ module Text.LLVM.Triple.AST
   ) where
 
 import Data.Data (Data)
-import Data.Typeable (Typeable)
```
kquick, line 24: I mentioned this elsewhere, but I'm not comfortable dropping `Typeable` yet, as there are known tools that are not yet compatible with the 3 latest GHC revisions that still use `llvm-pretty` and this has the risk of forcing them to upgrade before they are perhaps ready.  This is a pretty low level library, so it should lead on new version support and trail on dropping old versions.

langston-barrett: > Since GHC 7.10, all types automatically have [Typeable](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Data-Typeable.html#t:Typeable) instances derived.

https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Data-Typeable.html#t:Typeable

Do we really have tools that support GHC 7.8 (released Feb 2014)? I would be surprised if so, but happy to keep this if that's the case.

langston-barrett: It would also be worth investigating whether `llvm-pretty` compiles with 7.8 before these changes - I wouldn't be at all surprised if there were other sources of incompatibility already.

kquick: Oh, I assumed that `Typeable` was only automatic as of 9.6 or 9.8.  My concern is specifically 8.10, so if 8.10 builds with these `Typeable` instances removed then I retract my objections (it seems somewhat reasonable that anything building in the 8.x series should include support through 8.10, and 7 is definitely too old for any reasonable support).

langston-barrett: Makes sense since I bundled it with this upgrade 😄 I'll double-check to make sure it's clear in the commit message.

It feels a bit dodgy that we have projects that depend on this library but are unable to upgrade to a supported GHC version. Perhaps we should have a discussion about the range of versions we support?

It might also be nice to have a list of downstream projects, as is done [in parameterized-utils](https://github.com/GaloisInc/parameterized-utils/blob/9a65f3e76eeb84466834fe27d89730f473279d1e/doc/dev.md#downstream-projects). That would make it easier for contributors to `llvm-pretty` to be aware of the impacts of changes like this. If some of those projects are not open-source, we could perhaps list a point of contact and any relevant constraints (e.g., that a project must compile with the GHC distributed by the most recent LTS Ubuntu, or whatever).

langston-barrett:
I locally confirmed that the test suite still passes with GHC 8.10.7.

kquick (approved):

PR #166, opened 2025-10-27
Expose runBB for caller use

kquick:
The BB monad is used when generating blocks, and is customarily invoked via the `define` use of `defineBody` in the `DefineArgs` class.  However, there may be cases where the caller needs to wrap the BB monad, which `defineBody` does not support.  This exposes the `runBB` eliminator for the `BB` monad to allow those uses.

### DIFF
--- a/src/Text/LLVM.hs
+++ b/src/Text/LLVM.hs
@@ -40,6 +40,7 @@ module Text.LLVM (
 
     -- * Basic Blocks
   , BB()
+  , runBB
   , freshLabel
   , label
   , comment


src/Text/LLVM.hs
```diff
@@ -40,6 +40,7 @@ module Text.LLVM (
 
     -- * Basic Blocks
   , BB()
+  , runBB
```
RyanGlScott, line 43: Any reason not to just combine these into a single `BB(..)` export?

RyanGlScott (approved):

src/Text/LLVM.hs
kquick: This is the same convention used for `LLVM()` and `runLLVM` above.  I suppose both could be changed, but this way any future "private" portions of both types is preserved by default.  Not a strong argument, not a strong preference.

RyanGlScott: Fair enough. I'm fine with either convention, and doubly so since it's an already established convention.

PR #168, opened 2025-10-28
Add support for LLVM 19.

kquick:
This adds support for the new DebugRecords (https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/277) and GEP flag changes (https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/278 and https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/279).

This also makes changes to remove `icmp`, `fcmp` and `shl` constant expressions dropped in LLVM 19.

[See https://releases.llvm.org/19.1.0/docs/ReleaseNotes.html#changes-to-the-llvm-ir for more details.]

These changes are not backwards compatible.

These changes introduce new data structures, changes to existing data structures, and new semantics that may need to be handled downstream (there are associated changes to llvm-pretty-bc-parser).


kquick:
Needs CHANGELOG and cabal version updates prior to release.  Non-backwards compatible: major version number update required.

src/Text/LLVM/AST.hs
```diff
@@ -1408,6 +1419,63 @@ data FCmpOp = Ffalse  | Foeq | Fogt | Foge | Folt | Fole | Fone
     deriving (Data, Eq, Enum, Generic, Ord, Show, Typeable)
 
 
+-- Debug Instructions ----------------------------------------------------------
+
+data DebugRecord' lab
```
langston-barrett, line 1433: Might be nice to have a Haddock saying that these were added in LLVM 19?

src/Text/LLVM/DebugUtils.hs
```diff
@@ -365,20 +365,20 @@ localVariableNameDeclarations mdMap def =
   where
 
     aux :: [Stmt] -> Map Ident Ident -> Map Ident Ident
-    aux ( Effect (Store src dst _ _) _
-        : Effect (Call _ _ (ValSymbol (Symbol what)) [var,md,_]) _
+    aux ( Effect (Store src dst _ _) _ _
+        : Effect (Call _ _ (ValSymbol (Symbol what)) [var,md,_]) _ _
         : _) sofar
-      | what == "llvm.dbg.declare"
+      | what == "llvm.dbg.declare"  -- OLD: intrinsic declaration
```
langston-barrett: Might be nice to mention LLVM 19 here as well

src/Text/LLVM/AST.hs
```diff
@@ -1528,6 +1534,58 @@ data ConstExpr' lab
 
 type ConstExpr = ConstExpr' BlockLabel
 
+-- | Attributes imposing rules on the GEP; violating any rule results in a poison
+-- value.  If the base is a vector of pointers, the attributes apply to each
+-- computation element-wise.  See
+-- https://llvm.org/docs/LangRef.html#getelementptr-instruction for more
+-- information.
+data GEPOptionalFlag
```
RyanGlScott: I'm not crazy about the namely "`GEPOptionalFlag`", since the "`Optional`" part is implied by `GEP`/`ConstGEP` taking a list of these. I would be in favor of dropping the "`Optional`" part at the very least.

For that matter, how would you feel about naming this something like `GEPAttr`? The LLVM Language Reference calls these "attributes", so that would align more closely with the spec. Moreover, there is already some precedent for using "`Attr`" to name attribute-related data types in `llvm-pretty`. (`FunAttr`, `GlobalAttrs`, `DwarfAttrEncoding`, etc.)

src/Text/LLVM/AST.hs
```diff
@@ -1511,11 +1514,14 @@ extendMetadata md stmt = case stmt of
 -- Constant Expressions --------------------------------------------------------
 
 data ConstExpr' lab
-  = ConstGEP Bool (Maybe Word64) Type (Typed (Value' lab)) [Typed (Value' lab)]
+  = ConstGEP [GEPOptionalFlag] (Maybe RangeSpec) Type (Typed (Value' lab)) [Typed (Value' lab)]
```
RyanGlScott: It would be worth documenting the invariant that each unique type of GEP attribute can only appear at most once in this list. (This matters insofar as a `getelementptr` constant would be pretty-printed differently were an attribute to appear duplicated in this list.)

Similarly for the `GEP` data constructor above.

src/Text/LLVM/AST.hs
```diff
@@ -1528,6 +1534,58 @@ data ConstExpr' lab
...
 
+-- | Attributes imposing rules on the GEP; violating any rule results in a poison
+-- value.  If the base is a vector of pointers, the attributes apply to each
+-- computation element-wise.  See
+-- https://llvm.org/docs/LangRef.html#getelementptr-instruction for more
+-- information.
+data GEPOptionalFlag
+  = GEP_Inbounds
+    -- ^ Rules:
+    -- * Base pointer has an in_bounds (but not necessarily live) address of the
```
RyanGlScott: I'm not sure what "in_bounds" (note the underscore) is referring to, since nothing in the LLVM Language Reference uses that spelling. Should this just be "inbounds"?

src/Text/LLVM/AST.hs
```diff
@@ -1528,6 +1610,58 @@ data ConstExpr' lab
...
+    --   allocated object it is based on (i.e. points into that allocation or to
+    --   its end.  Size for a growable allocated object is the max size, not the
+    --   current size.
+    -- * Pointer must remain inbounds at all times when adding the offsets
+    -- * Implies NUSW
```
RyanGlScott: To make the connection more clear:

```suggestion
    -- * Implies 'GEP_NUSW'
```

src/Text/LLVM/AST.hs
```diff
@@ -1528,6 +1534,58 @@ data ConstExpr' lab
...
+
+orderedGEPOptionalFlags :: [GEPOptionalFlag]
+orderedGEPOptionalFlags = [GEP_Inbounds, GEP_NUSW, GEP_NUW] -- bit0, bit1, ...
+
+data RangeSpec
+  = RangeIndex Word64
+    -- ^ index of range
+  | Range Int Integer Integer
+    -- ^ width of arbitrary-precision integer (in bits) and lower and upper
+    -- arbitrary-precision integer bounds of that size
```
RyanGlScott: Worth documenting here that if you have `Range w start end`, then it denotes the range `[start, end)`, where `start` and `range` are `w`-bit integers.

langston-barrett (commented):
Thanks for putting this together! I just took a very high-level look and had some suggestions for comments to add.

src/Text/LLVM/AST.hs
```diff
@@ -1528,6 +1534,58 @@ data ConstExpr' lab
...
+    --   index type) and each offset (as unsigned) does not wrap the ptr
+    --   index type.
+    deriving (Data, Eq, Generic, Ord, Show, Typeable)
+
+orderedGEPOptionalFlags :: [GEPOptionalFlag]
+orderedGEPOptionalFlags = [GEP_Inbounds, GEP_NUSW, GEP_NUW] -- bit0, bit1, ...
+
+data RangeSpec
+  = RangeIndex Word64
+    -- ^ index of range
```
RyanGlScott: It might be worth documenting that `RangeIndex` exists solely to support an old version of the `getelementptr` constant where an individual index would be marked as `inrange`, and that modern versions of the `getelementptr` constant now use `Range` instead.

src/Text/LLVM/PP.hs
```diff
@@ -798,19 +798,20 @@ ppCallSym ty val = pp_ty <+> ppValue val
...
 
-ppGEP :: Bool -> Type -> Typed Value -> Fmt [Typed Value]
-ppGEP ib ty ptr ixs =
-  "getelementptr" <+> inbounds
+ppGEP :: [GEPOptionalFlag] -> Type -> Typed Value -> Fmt [Typed Value]
+ppGEP gf ty ptr ixs =
+  "getelementptr"
+    <+> (if inlineIsBool
+         then (if GEP_Inbounds `elem` gf then "inbounds" else empty)
+         else ppGepFlags gf)
```
RyanGlScott, line 815: Possibly silly question: what would go wrong if we simplified this code to just use `ppGepFlags gf` instead? It's my understanding that having `nusw` and `nuw` attributes on `getelementptr` constants is something that only happens with LLVM 19 or later, so my expectation is that such attributes wouldn't appear in the list in older LLVM versions.

src/Text/LLVM/PP.hs
```diff
@@ -978,6 +980,26 @@ ppConstExpr' pp expr =
         ppTupleT a b = parens $ ppTyped ppVal' a <> comma <+> ppTyp' b
         ppVal'       = ppValue' pp
         ppTyp'       = ppTyped ppVal'
+        ppRange =
+          let ppR = \case
+                RangeIndex i -> "index(" <+> integral i <+> ")" -- TODO FIXME
```
RyanGlScott: Do you plan to address this TODO?

My understanding (based on the grammar presented in [this old version of the LLVM Language Reference](https://releases.llvm.org/10.0.0/docs/LangRef.html#getelementptr-instruction)) is that the `inrange` attribute would have to be printed before each individual index, rather than at the front of the `getelementptr` constant as it is done here:

```
<result> = getelementptr inbounds <ty>, <ty>* <ptrval>{, [inrange] <ty> <idx>}*
```

test/Output.hs
```diff
@@ -167,6 +167,51 @@ tests = Tasty.testGroup "LLVM pretty-printing output tests"
       ----
       |]
 
+  , testCase "Stmt 1, LLVM 16" $
+    assertEqLines (ppToText $ ppLLVM 16 $ ppStmt s1) [sq|
+      In LLVM 11, DICompileUnit adds rangesBaseAddress, sysroot, and sdk
```
RyanGlScott: I'm not sure what the relevance of `DICompileUnit` is to `getelementptr`. Moreover, this mentions LLVM 11, but the pretty-printing is using LLVM 16.

test/Output.hs
```diff
@@ -167,6 +167,51 @@ tests = Tasty.testGroup "LLVM pretty-printing output tests"
...
+                                                        sdk: "SDK")
+      ----
+      |]
+
+  , testCase "Stmt 1, LLVM 17" $
+    -- LLVM 17 is notable in that the GEP instruction "inbounds" is no longer a
+    -- boolean but a flag with multiple possible values (only one is checked
+    -- here).
+    assertEqLines (ppToText $ ppLLVM 16 $ ppStmt s1) [sq|
+      In LLVM 11, DICompileUnit adds rangesBaseAddress, sysroot, and sdk
```
RyanGlScott: Similarly, comments about `DICompileUnit` here. Also, there is more inconsistency about LLVM versions here: LLVM 11, 16, and 17 are all mentioned. I think these should all be changed to use LLVM 17.

src/Text/LLVM/PP.hs
```diff
@@ -1008,6 +1039,68 @@ ppDebugInfo' pp di = case di of
   DebugInfoLabel dil            -> ppDILabel' pp dil
   DebugInfoArgList args         -> ppDIArgList' pp args
   DebugInfoAssignID             -> "!DIAssignID()"
+  -- DebugRecordDeclare drd        -> ppDbgRecDeclare' pp drd
+
+-- Prints DebugRecords (introduced in LLVM 19) which replace debug intrinsics and
+-- unlike the intrinsics that follow the instruction, the debug records *preceed*
```
RyanGlScott: Typofix:

```suggestion
-- unlike the intrinsics that follow the instruction, the debug records *precede*
```

src/Text/LLVM/AST.hs
```diff
@@ -1411,6 +1419,63 @@ data FCmpOp = Ffalse  | Foeq | Fogt | Foge | Folt | Fole | Fone
...
+  | DebugRecordValueSimple (DbgRecValueSimple' lab)
+  | DebugRecordLabel (DbgRecLabel' lab)
+  deriving (Data, Eq, Functor, Generic, Ord, Show, Typeable)
+
+data DbgRecValue' lab = DbgRecValue
+  {
+    drvLocation :: ValMd' lab -- DILocation
+  , drvLocalVariable :: ValMd' lab -- DILocalVariable
+  , drvExpression :: ValMd' lab -- DIExpression
+  , drvValAsMetadata :: ValMd' lab -- Metadata*
```
RyanGlScott: Perhaps these comments should be converted to Haddock comments? (Similarly for `DbgRecValueSimple'`, `DbgRecDeclare'`, `DbgRecAssign'`, and `DbgRecLabel'` below.)

src/Text/LLVM/AST.hs
```diff
@@ -1490,26 +1555,37 @@ elimValInteger _              = mzero
 -- Statements ------------------------------------------------------------------
 
 data Stmt' lab
-  = Result Ident (Instr' lab) [(String,ValMd' lab)]
-  | Effect (Instr' lab) [(String,ValMd' lab)]
+  = Result Ident (Instr' lab) [DebugRecord' lab] [(String, ValMd' lab)]
+  | Effect (Instr' lab) [DebugRecord' lab] [(String, ValMd' lab)]
```
RyanGlScott: It would be worth adding some Haddock comments to each of these `[DebugRecord' lab]` and `[(String, ValMd' lab)]` fields. My understanding is that (1) they both encode metadata information, but using different formats, and (2) it likely that you will only use one or the other depending on the LLVM version.

src/Text/LLVM/AST.hs
```diff
@@ -1490,26 +1555,37 @@ elimValInteger _              = mzero
...
+--
+-- Technically, DebugRecords are attached to Instructions, but since there's a
+-- 1:1 correspondence between Stmt and Instr, it is cleaner to attach the
+-- DebugRecords to the Stmt to keep the Instrs from getting additional
+-- complications.
 
-stmtMetadata :: Stmt' lab -> [(String,ValMd' lab)]
-stmtMetadata stmt = case stmt of
-  Result _ _ mds -> mds
-  Effect _ mds   -> mds
```
RyanGlScott, line 1503: Is there a particular reason to remove `stmtMetadata` outright? `llvm-pretty` still supports old-style debug intrinsics for the time being, so there may still be situations where one would want to extract the old-style metadata.

src/Text/LLVM/AST.hs
```diff
@@ -1672,6 +1747,10 @@ type DIFlags = Word32
 -- it stabilizes.
 type DIEmissionKind = Word8
 
+-- See https://github.com/llvm-mirror/llvm/blobl/release_?/include/llvm/BinaryFormat/Dwarf.h
+dwarf_DW_APPLE_ENUM_KIND_invalid :: Word64
+dwarf_DW_APPLE_ENUM_KIND_invalid = complement (0 :: Word64) -- LLVM 19
```
RyanGlScott: What is this used for?

src/Text/LLVM/AST.hs
```diff
@@ -1671,8 +1808,11 @@ data DICompositeType' lab = DICompositeType
   , dictAssociated     :: Maybe (ValMd' lab)
   , dictAllocated      :: Maybe (ValMd' lab)
   , dictRank           :: Maybe (ValMd' lab)
-  , dictAnnotations    :: Maybe (ValMd' lab)
-    -- ^ Introduced in LLVM 14.
+  , dictAnnotations    :: Maybe (ValMd' lab) -- ^ Introduced in LLVM 14.
+  , dictNumExtraInhabitants :: Word64        -- ^ added in LLVM 19.
```
RyanGlScott: Oh, I thought this was introduced in LLVM 20? At least, that's my reading of this LLVM commit: https://github.com/llvm/llvm-project/commit/f6617d65e496823c748236cdbe8e42bf4c8d8a55

src/Text/LLVM/PP.hs
RyanGlScott: Ah, now I see why you do this: it's because you add a `GEP_NUSW` attribute for old LLVM versions when parsing `getelementptr` constants (specifically, for `CeGepCode20`) in https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/316. And that's totally fair, since `GEP_Inbounds` is supposed to imply `GEP_NUSW`.

I retract this suggestion.

src/Text/LLVM/AST.hs
RyanGlScott: Ah, this is used in parsing `DICompositeType` values in https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/316.

src/Text/LLVM/AST.hs
kquick: Added.

src/Text/LLVM/AST.hs
kquick: Oops, that was from earlier work here.  Re-added.

src/Text/LLVM/PP.hs
kquick: That example is for the GEP instruction, not the ConstGEP.

I can confirm that the printing of the new LLVM-19  `Range` on the next line is correct via Ada-generated bitcode.

I have no examples that generate the old pre-LLVM-19 index value output, so the intention of this comment was to indicate that it was my best guess, but possibly wrong.  The other two options are to change it to throw an `error`, which is a bit harsher, or to emit more text that is obviously wrong and will therefore fail with anything that tries to interpret the .ll output.

Let me know what approach you think is best here.

src/Text/LLVM/AST.hs
kquick: When implementing these, I followed the general model for debug metadata, which was to specify them as `ValMd' lab` values.  The comments were to indicate what is expected, but in reality the consuming code will need to handle *any* `ValMd` value here, so making it a Haddock without more context might be misleading.

I'm tempted to use the actual DI record types expected here instead, but what that would mean is that if the parser read something unexpected, it would have to error out (unless these all became `Either` terms, which seems like a lot of overhead).  The LLVM implementation does a `cast` to the desired type, so it would presumably be comparable for us to use the more explicit type and fail on parsing, but that's a lot stricter than any of the other parsing.  Personally, I would prefer the stricter typing because that's much less work for the consumer (a `ValMd` is co-recursive with a `Value` and therefore infinitely variable), but perhaps we should address that (across the debug metadata) in a separate PR.

src/Text/LLVM/AST.hs
kquick: Fixed.

src/Text/LLVM/AST.hs
kquick: Updated.

src/Text/LLVM/AST.hs
kquick: Noted as a recommendation, and pretty-printing now nubs these to ensure it outputs correctly despite any duplicates present.

src/Text/LLVM/AST.hs
kquick: Not quite the right understanding: I moved some comments and enhanced them.  Let me know if you think more is needed here.

src/Text/LLVM/AST.hs
kquick: So it is.  I think I got this from the apple exaple, where it appeared earlier.  Updated the comments.

src/Text/LLVM/AST.hs
kquick: Updated.

src/Text/LLVM/AST.hs
kquick: Renamed.

src/Text/LLVM/DebugUtils.hs
kquick: Done.

src/Text/LLVM/AST.hs
RyanGlScott: I think using `ValMd` is fine.

To make this a bit more explicit, we could make the Haddock comments say something like "by convention, this is usually a `DIExpression` value" or some such. Would you be OK with that approach?

src/Text/LLVM/PP.hs
RyanGlScott: > I have no examples that generate the old pre-LLVM-19 index value output

An example of a `getelementptr` constant expression that uses `inrange` can be found [here](https://github.com/llvm/llvm-project/blob/d93620bf4d5e165999ad1c3cec2a2fdb02f56218/llvm/test/Bitcode/compatibility.ll#L1616-L1619):

```llvm
define i8** @constexpr() {
  ret i8** getelementptr inbounds ({ [4 x i8*], [4 x i8*] }, { [4 x i8*], [4 x i8*] }* null, i32 0, inrange i32 1, i32 2)
}
```

Is this enough to help you make the code above work on old LLVM versions?

src/Text/LLVM/AST.hs
kquick: That sounds fine (and done).

src/Text/LLVM/AST.hs
```diff
@@ -1408,6 +1423,68 @@ data FCmpOp = Ffalse  | Foeq | Fogt | Foge | Folt | Fole | Fone
...
 
 
+-- Debug Instructions ----------------------------------------------------------
+
+-- | Debug Instructions
+--
+-- In LLVM 19, debug instructions were added as a replacement for the intrinsic
+-- functions previously used.  This addition is described in
+-- llvm-project/llvm/docs/RemoveDIsDebugInfo.md in the LLVM repository.
+data DebugRecord' lab
```
RyanGlScott, line 1433: This should have a `type DebugRecord = DebugRecord' BlockLabel` type synonym for symmetry with the other definitions in this file. Similarly for `DbgRecValue'` and friends.

src/Text/LLVM/PP.hs
kquick: Yes, although that syntax is only valid in LLVM version prior to 19.  I've pushed an update to `llvm-pretty` pretty-printing to handle this case, and I've added a pre-llvm19 test to `llvm-pretty-bc-parser`'s `disasm-test` to verify this.

kquick:
All the issues have been addressed, and I have pending matching changes to `llvm-pretty-bc-parser`. 

The changes here are the bulk of the changes needed for LLVM 19.  Not yet supported: MMRA's, new/modified/renamed intrinsics, and pointer authority updates.

RyanGlScott (approved):

PR #169, opened 2025-10-29
Add flags to cast-related instructions (`ZExt`, `UiToFp`, and `Trunc`)

RyanGlScott:
Towards https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/306.


CHANGELOG.md
```diff
@@ -19,6 +19,13 @@
       unspecified bitwise representation.
     * Added `GoffMangling`, `WindowsX86CoffMangling`, and `XCoffMangling` forms
       of Mangling.
+  * Changes to cast-related instructions:
+    * Add a `Bool` field to `ZExt` which, if `True`, indicates that the
+      argument must be non-negative.
+    * Add a `Bool` field to `UiToFp` which, if `True`, indicates that the
+      argument must be non-negative.
```
RyanGlScott:
```suggestion
    * Add a `Bool` field to `ZExt` which, if `True`, which indicates that the
      argument must be non-negative.
    * Add a `Bool` field to `UiToFp` which, if `True`, which indicates that the
      argument must be non-negative.
```

langston-barrett (approved):

CHANGELOG.md
kquick: Too many whiches in the suggested change.

src/Text/LLVM/AST.hs
```diff
@@ -1061,19 +1061,51 @@ data BitOp
...
-  = Trunc
-  | ZExt
+  = Trunc Bool Bool
+    -- ^ Truncate an integer value to a smaller integer type.
+    --
+    -- The 'Bool' fields (added in in LLVM 20) encode whether to perform
+    -- overflow-related checks:
+    --
+    -- * First 'Bool': check for unsigned overflow.
+    -- * Second 'Bool': check for signed overflow.
```
kquick: Do these added bools have the same semantics as https://github.com/GaloisInc/llvm-pretty/pull/168/files#diff-c60254e336a88f05888286b397c8ba49f578b01ce4ce3cff4a1d6431008a89e2R1652-R1675?  I wonder if those flags should be extracted and generalized to use in both locations?

I'll note that the two flags seem mutually exclusive to me (both here and in the reference location above), and it bothers me that they could thus both be set to true.  I think this may be an artifact of how they need to be referenced in the bitcode file but I also wonder if you agree about their exclusivity and if we should thus use a trinary value here instead of two individual and independent flags?

CHANGELOG.md
RyanGlScott: Wow, that's what I get for trying to write documentation first thing in the morning. I'll retract this suggestion.

kquick (approved):

PR #172, opened 2025-11-15
Update pretty-printing for items dropped in LLVM 17 and LLVM 18.

kquick:


src/Text/LLVM/PP.hs
```diff
@@ -980,9 +980,24 @@ ppConstExpr' pp expr =
...
+    ConstConv op tv t  ->
+      let droppedIn18 = case op of
+                          ZExt _ -> True
+                          SExt -> True
+                          FpTrunc -> True
+                          FpExt -> True
+                          FpToUi -> True
+                          FpToSi -> True
+                          UiToFp _ -> True
+                          SiToFp -> True
```
RyanGlScott, line 994: It might be helpful to cite the relevant LLVM project commit that removed these as constexprs—I believe that's https://github.com/llvm/llvm-project/commit/17764d2c87bad3b9654b7310c9936c0d15e9bf96.

src/Text/LLVM/PP.hs
```diff
@@ -980,9 +980,24 @@ ppConstExpr' pp expr =
                        in reverse  -- ppTyp's pushes entries to the listg head
                           $ foldl (ppTyp's mrng) [ppType ty]
                           $ zip argIndices (ptr:ixs)))
-    ConstConv op tv t  -> ppConvOp op <+> parens (ppTyp' tv <+> "to" <+> ppType t)
+    ConstConv op tv t  ->
+      let droppedIn18 = case op of
+                          ZExt _ -> True
+                          SExt -> True
```
RyanGlScott, line 987: These two were removed in https://github.com/llvm/llvm-project/commit/e4a4122eb6768b69640173b4c32fd88de4547227.

src/Text/LLVM/PP.hs
```diff
@@ -980,9 +980,24 @@ ppConstExpr' pp expr =
...
+                          UiToFp _ -> True
+                          SiToFp -> True
+                          _ -> False
+          ppConstConv = ppConvOp op <+> parens (ppTyp' tv <+> "to" <+> ppType t)
+      in if droppedIn18
+         then droppedInLLVM 18 "fptrunc/fpext/fptoui/fptosi/uitofp/sitofp constexprs" ppConstConv
+         else ppConstConv
     ConstSelect c l r  ->
-      "select" <+> parens (commas [ ppTyp' c, ppTyp' l , ppTyp' r])
+      droppedInLLVM 17 "select constexpr"
```
RyanGlScott: This one was removed in https://github.com/llvm/llvm-project/commit/bbfb13a5ffbccf1759ca6b75262a3ffdbe20496e.

src/Text/LLVM/PP.hs
RyanGlScott, line 1008: These two were dropped in https://github.com/llvm/llvm-project/commit/63dc31b68b78bc0e5deef21b98cab72de997c471.

src/Text/LLVM/PP.hs
```diff
@@ -992,7 +1007,9 @@ ppConstExpr' pp expr =
     ConstUnaryArith op a   -> ppUnaryArithOp op <+> ppTyp' a
     ConstBit        op@(Shl _ _) a b -> droppedInLLVM 19 "shl constexprs"
```
RyanGlScott, line 1011: This one was dropped in https://github.com/llvm/llvm-project/commit/76d3ab2cc33336c1eece6484c9a55577eac7e79a.

src/Text/LLVM/PP.hs
```diff
@@ -992,7 +1007,9 @@ ppConstExpr' pp expr =
...
                                         $ ppBitOp op   <+> ppTuple a b
-    ConstBit        op a b -> ppBitOp op   <+> ppTuple a b
+    ConstBit        Xor a b -> ppBitOp Xor <+> ppTuple a b
+    ConstBit        op a b -> droppedInLLVM 18 "and/or/lshr/ashr constexprs"
+                              $ ppBitOp op <+> ppTuple a b
```
RyanGlScott, line 1015: Rather than special-casing `Xor`, how about defining something like this for symmetry with the code in the `ConstConv` case?

```hs
ConstBit op a b ->
  let droppedIn18 =
        case op of
          -- Removed in https://github.com/llvm/llvm-project/commit/625113402f9febd4d8c907a342ea09a3c0982ba8
          And -> True
          Or -> True
          -- Removed in https://github.com/llvm/llvm-project/commit/56c1d30183e156365f7057f5945b2bc48fdb32e7
          Lshr -> True
          Ashr -> True
          Xor -> False in ...
```

RyanGlScott (approved):

src/Text/LLVM/PP.hs
kquick: Given that `Xor` is the only one left that is supported, and I expect it to be removed soon, it will need a different "droppedInLLVM invocation anyhow (like `Shl`).

PR #175, opened 2025-11-17
Make `dictEnumKind` and `dwarf_DW_APPLE_ENUM_KIND` 32 (not 64) bits

### LINKED ISSUE #173: `dwarf_DW_APPLE_ENUM_KIND_invalid` has the wrong value
Per the [LLVM source code](https://github.com/llvm/llvm-project/blob/eb9d56cb558cd7412edc8488f331517f3bb3df30/llvm/include/llvm/BinaryFormat/Dwarf.h#L51), `DW_APPLE_ENUM_KIND_invalid` is intended to be a 32-bit unsigned integer with the value `~0U`, i.e., where all bits are set. `dwarf_DW_APPLE_ENUM_KIND_invalid` (the `llvm-pretty` counterpart to this value), on the other hand, is defined as a _64-bit_ unsigned integer value where all bits are set:

https://github.com/GaloisInc/llvm-pretty/blob/5e45daf9c2b8fd094962929a195b6cf2010738cf/src/Text/LLVM/AST.hs#L1825-L1826

This means that `dwarf_DW_APPLE_ENUM_KIND_invalid` incorrectly has the value `18446744073709551615` instead of `4294967295` (as intended). This matters for `llvm-pretty-bc-parser` purposes, as `dictEnumKind` values have to be compared against the value of `dwarf_DW_APPLE_ENUM_KIND_invalid` during parsing, as seen in the code [here](https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/baf33c199c61c6f15ea7d47cd71278b66a4fca68/src/Data/LLVM/BitCode/IR/Metadata.hs#L693-L698). When running `llvm-pretty-bc-parser`'s `disasm-test` test suite using LLVM 21, this results in round-tripping errors that look like:

```
llvm-as: error: value for 'enumKind' too large, limit is 1
  runtimeLang: 0, enumKind: 4294967295)
                            ^
```

To fix this, I propose that we change the type of `dictEnumKind` and `dwarf_DW_APPLE_ENUM_KIND_invalid` to use `Word32` instead of `Word64` to match LLVM.

RyanGlScott:
Fixes #173.

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -1822,8 +1822,8 @@ type DIFlags = Word32
 type DIEmissionKind = Word8
 
 -- See https://github.com/llvm/llvm-project/commit/eb8901bda11fd55deeecd067fc4c9dcc0fb89984
-dwarf_DW_APPLE_ENUM_KIND_invalid :: Word64
-dwarf_DW_APPLE_ENUM_KIND_invalid = complement (0 :: Word64) -- ~ LLVM 19
+dwarf_DW_APPLE_ENUM_KIND_invalid :: Word32
+dwarf_DW_APPLE_ENUM_KIND_invalid = complement (0 :: Word32) -- ~ LLVM 19
 
 data DIBasicType = DIBasicType
   { dibtTag      :: DwarfTag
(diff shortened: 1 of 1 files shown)


langston-barrett (approved):

langston-barrett:
Does this deserve a CHANGELOG entry?

RyanGlScott:
This bugfix only affects `dictEnumKind`/`dwarf_DW_APPLE_ENUM_KIND`, which hasn't yet appeared in a Hackage release. I don't think there's anything we need to note here.

kquick (approved):

PR #178, opened 2025-12-07
Add checksumkind and checksum to DIFile and parameterize tick version

kquick:


src/Text/LLVM/AST.hs
```diff
@@ -1951,10 +1952,22 @@ data DIExpression = DIExpression
   { dieElements :: [Word64]
   } deriving (Data, Eq, Generic, Ord, Show, Typeable)
 
-data DIFile = DIFile
+data DIFile' lab = DIFile
   { difFilename  :: FilePath
   , difDirectory :: FilePath
-  } deriving (Data, Eq, Generic, Ord, Show, Typeable)
+  , difChecksumKind :: DIChecksumKind
+  , difChecksum :: Maybe String
```
RyanGlScott, line 1959: Were these fields introduced in a particular LLVM version? If so, should this be documented here in the Haddocks?

PR #179, opened 2025-12-07
Add bbStmtModifier to allow generated statements to be updated (e.g. …

kquick:
…to add debug metadata)

### DIFF
--- a/src/Text/LLVM.hs
+++ b/src/Text/LLVM.hs
@@ -41,6 +41,7 @@ module Text.LLVM (
     -- * Basic Blocks
   , BB()
   , runBB
+  , bbStmtModifier
   , freshLabel
   , label
   , comment
@@ -314,8 +315,11 @@ define' attrs rty sym sig va k = do
 -- Basic Block Monad -----------------------------------------------------------
 
 newtype BB a = BB
-  { unBB :: WriterT [BasicBlock] (StateT RW Id) a
+  { unBB :: ReaderT (Stmt -> Stmt) (WriterT [BasicBlock] (StateT RW Id)) a
   } deriving (Functor,Applicative,Monad,MonadFix)
+
+bbStmtModifier :: (Stmt -> Stmt) -> BB a -> BB a
+bbStmtModifier stmtModifier = BB . local stmtModifier . unBB
 
 avoidName :: String -> BB ()
 avoidName name = BB $ do
@@ -333,7 +337,7 @@ freshNameBB pfx = BB $ do
 
 runBB :: BB a -> (a,[BasicBlock])
 runBB m =
-  case runId (runStateT emptyRW (runWriterT (unBB body))) of
+  case runId (runStateT emptyRW (runWriterT (runReaderT id (unBB body)))) of
     ((a,bbs),_rw) -> (a,bbs)
   where
   -- make sure that the last block is terminated
@@ -367,7 +371,8 @@ emitStmt :: Stmt -> BB ()
 emitStmt stmt = do
   BB $ do
     rw <- get
-    set $! rw { rwStmts = rwStmts rw Seq.|> stmt }
+    smod <- ask
+    set $! rw { rwStmts = rwStmts rw Seq.|> smod stmt }
   when (isTerminator (stmtInstr stmt)) terminateBasicBlock
 
 effect :: Instr -> BB ()
--- a/src/Text/LLVM.hs
+++ b/src/Text/LLVM.hs
@@ -319,7 +319,7 @@ newtype BB a = BB
   } deriving (Functor,Applicative,Monad,MonadFix)
 
 
--- | The 'bbStmtModifier' can be used to register a function that can be used to
+-- | The 'bbStmtModifier' function can be used to register a function that can
 -- modify the subsequent statements generated into this block.
 --
 -- For example the following @BB@ monad code segment will emit a couple of LLVM
--- a/src/Text/LLVM.hs
+++ b/src/Text/LLVM.hs
@@ -329,7 +329,7 @@ newtype BB a = BB
 --    call fooFunc [v]
 --    jump end
 --
--- but these statements will be "plain" in the resulting @BasicBlock@.  If the
+-- But these statements will be \"plain\" in the resulting 'BasicBlock'.  If the
 -- caller wishes to add debug Metadata for location, they could instead write:
 --
 --    bbStmtModifier (extendMetadata ("dbg", ValMdRef i))
--- a/src/Text/LLVM.hs
+++ b/src/Text/LLVM.hs
@@ -325,9 +325,9 @@ newtype BB a = BB
 -- For example, the following 'BB' monad code segment will emit a couple of LLVM
 -- statements:
 --
---    v <- load (iT 8) globalVar Nothing
---    call fooFunc [v]
---    jump end
+-- > v <- load (iT 8) globalVar Nothing
+-- > call fooFunc [v]
+-- > jump end
 --
 -- But these statements will be \"plain\" in the resulting 'BasicBlock'.  If the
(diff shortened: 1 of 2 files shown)


RyanGlScott (commented):
I think I'm missing some context here. What is the purpose of this new `bbStmtModifier` function? Where is it used? (I don't see any call sites.)

kquick:
If you use `llvm-pretty` functions from the top-level `LLVM.hs` file to generate instructions (e.g. `ret`, `assign`, `jump`) they all operate in the `BB` monad, which use the internal `emitStmt` function to add the instructions to a Writer.  Those operations only allow specification of the primary instruction arguments, but not any metadata that might be associated with the function.  

This PR adds a `Reader` context to `BB` which provides a function that can modify the generated stmt before it is appended to the `Writer`; the caller can use this to affect the generated statement in any way it wishes, including adding debug Metadata.

There are no callers to this in `llvm-pretty` itself, and the default stmt modifier function is the identity function, but this can be used by clients of `llvm-pretty`.

RyanGlScott:
> There are no callers to this in `llvm-pretty` itself, and the default stmt modifier function is the identity function, but this can be used by clients of `llvm-pretty`.

By "clients of `llvm-pretty`", I assume you mean `llvm-pretty-bc-parser` here? Could you give an example of how this function will be used in `llvm-pretty-bc-parser`?

I'm not opposed to adding something like this in principle, although I do wonder if carrying around a giant closure of type `Stmt -> Stmt` is more powerful than what we actually need.

kquick:
Nope, `llvm-pretty-bc-parser` is not expected to use this because its parser directly generates the AST rather than using these functions to *generate* statements.

This is being used in a separate project where LLVM IR is being synthesized via `llvm-pretty`.

While the current primary use case is adding debug Metadata, the `Stmt -> Stmt` processing seemed to be the most flexible option.  Referencing a function seemed to me to be fairly idiomatic Haskell, and it was not my understanding that `id` would be a "giant closure"; I would expect the closure size to be dependent on the client override of this element (if any), but maybe I'm misunderstanding the "size" aspect of a function pointer like this?   I'm also open to other ideas on how this capability might be implemented.

RyanGlScott:
> This is being used in a separate project where LLVM IR is being synthesized via `llvm-pretty`.

I stand corrected! I didn't realize that we were using `llvm-pretty` elsewhere.

> the current primary use case is adding debug Metadata

Thanks, that's helpful to know. That seems like a pretty reasonable use case, and I wonder if we should restrict the type of `bbStmtModifier` to just reflect this. Having the ability to arbitrarily change the `Stmt` is a lot of power, and I worry that there may be consequences for doing things like changing a non-terminator `Stmt` to a terminator `Stmt`.

> it was not my understanding that `id` would be a "giant closure"

It's not `id` that would large, but rather it would hinge on whatever you pass in as a higher-order function to `bbStmtModifier`. If the function closes over many different values, then that could potentially result in a large closure. (I'm speculating since I haven't seen the call site for this function.)

If the use case is, say, adding a debug record to a particular `Stmt`, then I think we could instead add a function with a type like `:: DebugRecord -> BB a -> BB a`, which would avoid this issue altogether.

kquick:
My concern with your proposal is that the `BB` accumulates a number of statements, so it's not clear which ones the `DebugRecord` should apply to.  With the `Stmt -> Stmt` approach it is more discrete and the caller can update the function before passing additional statements to be aggregated into the `BB`.  I agree it's more dangerous, but given that you can create a `BB` using these functions that does not have a terminating statement, I don't think it's any more dangerous than existing functionality.

RyanGlScott:
> My concern with your proposal is that the `BB` accumulates a number of statements, so it's not clear which ones the `DebugRecord` should apply to.

That's fair. But then again, I think the same concern applies to the `Stmt -> Stmt` approach—which `Stmt`s in particular do the function get applied to?

I think I can live with this approach, but we should definitely add some Haddocks to the `bbStmtModifier` function stating precisely which `Stmt`s are modified, along with some examples of how it is meant to be invoked.

src/Text/LLVM.hs
```diff
@@ -314,9 +315,36 @@ define' attrs rty sym sig va k = do
 -- Basic Block Monad -----------------------------------------------------------
 
 newtype BB a = BB
-  { unBB :: WriterT [BasicBlock] (StateT RW Id) a
+  { unBB :: ReaderT (Stmt -> Stmt) (WriterT [BasicBlock] (StateT RW Id)) a
   } deriving (Functor,Applicative,Monad,MonadFix)
 
+
+-- | The 'bbStmtModifier' can be used to register a function that can be used to
+-- modify the subsequent statements generated into this block.
```
RyanGlScott: Copy-editing:


```suggestion
-- | The 'bbStmtModifier' function can be used to register a function that can
-- modify the subsequent statements generated into this block.
```

The phrase "can be used to" is used twice in this sentence, so I opted to delete the second one.

src/Text/LLVM.hs
```diff
@@ -314,9 +315,36 @@ define' attrs rty sym sig va k = do
...
+--
+-- For example the following @BB@ monad code segment will emit a couple of LLVM
+-- statements:
```
RyanGlScott:
```suggestion
-- For example, the following 'BB' monad code segment will emit a couple of LLVM
-- statements:
```

src/Text/LLVM.hs
```diff
@@ -314,9 +315,36 @@ define' attrs rty sym sig va k = do
...
+--
+--    v <- load (iT 8) globalVar Nothing
+--    call fooFunc [v]
+--    jump end
```
RyanGlScott: Use Haddock syntax for code blocks:


```suggestion
-- > v <- load (iT 8) globalVar Nothing
-- > call fooFunc [v]
-- > jump end
```

src/Text/LLVM.hs
```diff
@@ -314,9 +315,36 @@ define' attrs rty sym sig va k = do
...
+--
+-- but these statements will be "plain" in the resulting @BasicBlock@.  If the
+-- caller wishes to add debug Metadata for location, they could instead write:
```
RyanGlScott:
```suggestion
-- But these statements will be \"plain\" in the resulting 'BasicBlock'.  If the
-- caller wishes to add debug Metadata for location, they could instead write:
```

src/Text/LLVM.hs
```diff
@@ -314,9 +315,36 @@ define' attrs rty sym sig va k = do
...
+--
+--    bbStmtModifier (extendMetadata ("dbg", ValMdRef i))
+--    v <- load (iT 8) globalVar Nothing
+--    bbStmtModifier (extendMetadata ("dbg", ValMdRef j))
+--    call fooFunc [v]
+--    jump end
```
RyanGlScott:
```suggestion
-- > bbStmtModifier (extendMetadata ("dbg", ValMdRef i))
-- >  v <- load (iT 8) globalVar Nothing
-- > bbStmtModifier (extendMetadata ("dbg", ValMdRef j))
-- > call fooFunc [v]
-- > jump end
```

src/Text/LLVM.hs
```diff
@@ -314,9 +315,36 @@ define' attrs rty sym sig va k = do
...
+--
+-- Where @i@ and @j@ are the metadata index values of the @DebugLoc@ entries
+-- describing the source location of the "load" and "call"+"jump" statements,
+-- respectively.
```
RyanGlScott:
```suggestion
-- Where @i@ and @j@ are the metadata index values of the 'DebugLoc' entries
-- describing the source location of the \"load\" and \"call\"+\"jump\" statements,
-- respectively.
```

RyanGlScott (approved):
Worth mentioning this new API entrypoint in the changelog, but otherwise this LGTM.

PR #180, opened 2025-12-07
Add enumeration values for DwarfVirtuality, DIFlags, and DIEmissionKind

kquick:


src/Text/LLVM/PP.hs
```diff
@@ -1215,8 +1215,7 @@ ppDIBasicType' pp bt = "!DIBasicType"
        ,     (("size:"     <+>) . ppSizeOrOffsetValMd' pp) <$> dibtSize bt
        , pure ("align:"    <+> integral (dibtAlign bt))
        , pure ("encoding:" <+> integral (dibtEncoding bt))
-       ,     (("flags:"    <+>) . integral)
-             <$> dibtFlags bt
+       , (\f -> "flags:" <+> (commas (text . show <$> f))) <$> dibtFlags bt
```
RyanGlScott, line 1218: To my knowledge, this is the first time that we are relying on the output of a derived `Show` instance in the pretty-printing output. I think that this is fine, but we should add some comments to the relevant data type(s) saying that the names of the data constructors must match what is in LLVM _exactly_, as the pretty-printing code depends on this property.

src/Text/LLVM/AST.hs
```diff
@@ -1818,13 +1821,93 @@ data DINameSpace' lab = DINameSpace
...
--- differently across versions. Maybe turn this into a sum type once
--- it stabilizes.
-type DIEmissionKind = Word8
+
+type DwarfVirtuality = DISPFlags
+-- | See
+-- llvm-project/llvm/include/llvm/IR/{DebugInfoMetadata.h,DebugInfoFlags.def}
+type DISPFlags = [DISPFlag]
+data DISPFlag
+  = DISPFlagNonvirtual -- DISPFlagZero -- == 0
```
RyanGlScott, line 1830: Where is `DISPFlagNonvirtual` defined on the LLVM side? If I'm reading [this](https://github.com/llvm/llvm-project/blob/07bafab83de2ba9f7d15fb194e0d34cd71f17af2/llvm/include/llvm/IR/DebugInfoFlags.def#L76-L78) correctly, shouldn't it be named `DISPFlagZero`?

src/Text/LLVM/PP.hs
```diff
@@ -1424,9 +1423,9 @@ ppDISubprogram' pp sp = "!DISubprogram"
        , pure ("isDefinition:"    <+> ppBool (dispIsDefinition sp))
        , pure ("scopeLine:"       <+> integral (dispScopeLine sp))
        ,      (("containingType:" <+>) . ppValMd' pp) <$> (dispContainingType sp)
-       , pure ("virtuality:"      <+> integral (dispVirtuality sp))
```
RyanGlScott, line 1427: The `virtuality` field is now being displayed as `spflags`. Is this a change that was introduced in a particular LLVM version? Does the pretty-printing code need to do different things depending on the LLVM version?

src/Text/LLVM/PP.hs
```diff
@@ -1274,7 +1273,7 @@ ppDICompositeType' pp ct = "!DICompositeType"
        ,     (("size:"           <+>) . ppSizeOrOffsetValMd' pp) <$> dictSize ct
        , pure ("align:"          <+> integral (dictAlign ct))
        ,     (("offset:"         <+>) . ppSizeOrOffsetValMd' pp) <$> dictOffset ct
-       , pure ("flags:"          <+> integral (dictFlags ct))
+       , pure ("flags:"          <+> (commas (text . show <$> (dictFlags ct))))
```
RyanGlScott, line 1276: These parentheses are redundant:


```suggestion
       , pure ("flags:"          <+> (commas (text . show <$> dictFlags ct)))
```

Similar comments apply to other uses of `text . show <$> (...)` introduced in this PR.

RyanGlScott:
NB: This PR would fix https://github.com/GaloisInc/llvm-pretty/issues/32.

PR #182, opened 2026-01-22
Prepare release 0.14.0.0

aschwerdfeger-galois:


RyanGlScott:
Regarding the CI failure in the GHC 9.12.3 job, upstream `ghcup` has [marked 9.12.3 as broken](https://github.com/haskell/ghcup-metadata/commit/fcc424505ed13af5241bf00da2414a9588657f65) (see [this announcement](https://discourse.haskell.org/t/critical-code-generation-bug-with-ghc-9-12-3/13505)), which causes `ghcup` to no longer offer it by default unless you take extra steps to enable it. I think that we should just downgrade to 9.12.2 for the time being.

aschwerdfeger-galois:
If there are no objections, I will close this PR and delete the branch, then recreate it after the GHC downgrade.

RyanGlScott:
That's fine, although I would also be OK with keeping this PR open and simply rebasing it on top of the GHC-related changes after they've landed. Alternatively, we could force-push the GHC-related commits before the `llvm-pretty` version bump commits—your call.

aschwerdfeger-galois:
Closing this PR to recreate the release branch on top of CI fixes.

PR #186, opened 2026-01-29
Fill in missing cases for the float types in "FloatType"

### LINKED ISSUE #66: Improve support for Float types
There is currently support lacking for all of the different Float types (e.g.  at https://github.com/elliottt/llvm-pretty/blob/master/src/Text/LLVM/AST.hs#L346).  Work should be done to add more complete FloatType support.

sauclovian-g:
Closes #66.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -28,7 +28,7 @@ jobs:
     uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v1
     with:
       cabal: ${{ matrix.cabal }}
-      cache-key-prefix: v2
+      cache-key-prefix: v3a
       ghc: ${{ matrix.ghc }}
       os: ${{ matrix.os }}
       test: ${{ matrix.test }}
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -25,7 +25,7 @@ jobs:
             build-targets: "lib:llvm-pretty"
       fail-fast: false
     name: ${{ matrix.os }} - GHC ${{ matrix.ghc }} - Cabal ${{ matrix.cabal }}
-    uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v1
+    uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v2
     with:
       cabal: ${{ matrix.cabal }}
       cache-key-prefix: v3a
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -125,7 +125,11 @@ module Text.LLVM.AST
   , FCmpOp(..)
     -- * Values
   , Value'(..), Value
+  , FPHalfValue(..)
+  , FPBFloatValue(..)
   , FP80Value(..)
+  , FP128Value(..)
+  , FP128_PPCValue(..)
   , ValMd'(..), ValMd
   , KindMd
   , FnMdAttachments
@@ -507,6 +511,7 @@ data PrimType
 
 data FloatType
   = Half
+  | BFloat -- ^ Introduced in LLVM 11
   | Float
   | Double
   | Fp128
@@ -713,10 +718,13 @@ primTypeNull (FloatType ft) = floatTypeNull ft
 primTypeNull _              = ValZeroInit
 
 floatTypeNull :: FloatType -> Value' lab
+floatTypeNull Half     = ValHalf $ FPHalf 0
+floatTypeNull BFloat   = ValBFloat $ FPBFloat 0
 floatTypeNull Float    = ValFloat 0
-floatTypeNull Double   = ValDouble 0 -- XXX not sure about this
+floatTypeNull Double   = ValDouble 0
+floatTypeNull Fp128    = ValFP128 $ FP128_LongDouble 0 0
 floatTypeNull X86_fp80 = ValFP80 $ FP80_LongDouble 0 0
-floatTypeNull _        = error "must be a float type"
+floatTypeNull PPC_fp128 = ValFP128_PPC $ FP128_PPC_DoubleDouble 0 0
 
 typeNull :: Type -> NullResult lab
 typeNull (PrimType pt) = HasNull (primTypeNull pt)
@@ -1538,9 +1546,13 @@ type DbgRecLabel = DbgRecLabel' BlockLabel
 data Value' lab
   = ValInteger Integer
   | ValBool Bool
+  | ValHalf FPHalfValue
+  | ValBFloat FPBFloatValue
   | ValFloat Float
   | ValDouble Double
   | ValFP80 FP80Value
+  | ValFP128 FP128Value
+  | ValFP128_PPC FP128_PPCValue
   | ValIdent Ident
   | ValSymbol Symbol
   | ValNull
@@ -1560,7 +1572,30 @@ data Value' lab
 
 type Value = Value' BlockLabel
 
+-- | 16-bit half-precision floating point value (IEEE half)
+data FPHalfValue = FPHalf Word16
+               deriving (Data, Eq, Ord, Generic, Show)
+
+-- | Different 16-bit half-precision floating point value
+--   ("Brain" or "bfloat16")
+data FPBFloatValue = FPBFloat Word16
+               deriving (Data, Eq, Ord, Generic, Show)
+
+-- | x86 80-bit long double floating point value
+--   (note that there's also an m86k 80-bit float that's almost but
+--   not quite the same)
 data FP80Value = FP80_LongDouble Word16 Word64
+               deriving (Data, Eq, Ord, Generic, Show)
+
+-- | IEEE quad-precision long-double floating point value
+data FP128Value = FP128_LongDouble Word64 Word64
+               deriving (Data, Eq, Ord, Generic, Show)
+
+-- | PowerPC pair-of-doubles floating point value
+--   (The value represented is the sum of the two doubles, which
+--   normally but not necessarily have exponents chosen so this makes
+--   sense.)
+data FP128_PPCValue = FP128_PPC_DoubleDouble Double Double
                deriving (Data, Eq, Ord, Generic, Show)
 
 data ValMd' lab
@@ -1593,9 +1628,13 @@ type DebugLoc = DebugLoc' BlockLabel
 isConst :: Value' lab -> Bool
 isConst ValInteger{}   = True
 isConst ValBool{}      = True
+isConst ValBFloat{}    = True
+isConst ValHalf{}      = True
 isConst ValFloat{}     = True
 isConst ValDouble{}    = True
 isConst ValFP80{}      = True
+isConst ValFP128{}     = True
+isConst ValFP128_PPC{} = True
 isConst ValConstExpr{} = True
 isConst ValZeroInit    = True
 isConst ValNull        = True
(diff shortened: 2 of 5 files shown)


src/Text/LLVM/PP.hs
```diff
@@ -865,11 +866,38 @@ ppFCmpOp Fune   = "une"
...
 ppValue' :: Fmt i -> Fmt (Value' i)
 ppValue' pp val = case val of
   ValInteger i       -> integer i
   ValBool b          -> ppBool b
   -- Note: for +Inf/-Inf/NaNs, we want to output the bit-correct sequence
+  ValHalf (FPHalf x) ->
+    -- Half floats have a 1-bit sign, 5-bit exponent, and 10-bit
+    -- significand.
+    let expbits = x .&. 0x7c00 in
+    if expbits == 0x7c00 -- isInfinite || isNan
```
RyanGlScott: Should I interpret the `-- isInfinite || isNan` comment here as "this expression will be true if `x` (represented as a half float) is infinite or a NaN value"?

src/Text/LLVM/PP.hs
```diff
@@ -865,11 +866,38 @@ ppFCmpOp Fune   = "une"
...
+      then text "0xH" <> text (zerofill 4 $ showHex x "")
+      else
+        let -- Convert to 32-bit single float (1-bit sign, 8-bit exponent,
+            -- 23-bit significand)
+            sign = shiftL (fromIntegral (x .&. 0x8000)) (31 - 15)
+            signif = shiftL (fromIntegral (x .&. 0xfff)) (23 - 10)
+            expo = case fromIntegral (shiftR expbits 10) of
+                0 -> 0
+                (n :: Int) -> shiftL (fromIntegral ((n - 15) + 127)) 23
```
RyanGlScott: Minor style nit pick: I would prefer to state that we're converting to an `Int` nearer to the call to `fromIntegral`, rather than in the `n :: Int` pattern some distance away:

```suggestion
            expo = case fromIntegral @Word @Int (shiftR expbits 10) of
                0 -> 0
                n -> shiftL (fromIntegral ((n - 15) + 127)) 23
```

(Generally speaking, I try to always use explicit type applications with every call to `fromIntegral`, as it's all too easy for the underlying types to change without one's knowledge when refactoring code, which can cause subtle logic errors later on.)

RyanGlScott (commented):
Please make sure to document these additions in the changelog. I expect that the new `Value'` constructors will require some changes on the `llvm-pretty-bc-parser` and `crucible-llvm` side (although they will likely be very mechanical changes).

src/Text/LLVM/PP.hs
sauclovian-g: Yes. If you have a better wording, please fire away

RyanGlScott: How about "`-- True if x is infinite or a NaN value`"?

src/Text/LLVM/PP.hs
sauclovian-g: `fromIntegral` seems to be the only way to deal with `Word16`. I think I ended up putting the type there because it was the least ugly place to stuff a type signature, and I don't think I even knew about `TypeApplications`.

Here I think the in and out types are adequately nailed down ( nd not subject to future change) that we only really need the intermediate `Int`.

Anyway, fixed.

src/Text/LLVM/AST.hs
```diff
@@ -507,6 +511,7 @@ data PrimType
 
 data FloatType
   = Half
+  | BFloat
```
RyanGlScott: Worth adding a comment here that this is only available in LLVM 11 or later:


```suggestion
  | BFloat -- ^ Introduced in LLVM 11
```

sauclovian-g: Indeed, and I wouldn't have known where to find that.

sauclovian-g:
I have changes for llvm-pretty-bc-parser and minimal changes for crucible. Adding bfloat support in Crucible requires adding it to what4, which looks not entirely trivial even absent the complications described in https://github.com/GaloisInc/what4/issues/342.

src/Text/LLVM/PP.hs
sauclovian-g: Hmm I think GH just didn't show me this earlier. Anyway, that's better, I'll change it

kquick (approved):

src/Text/LLVM/PP.hs
```diff
@@ -865,11 +868,52 @@ ppFCmpOp Fune   = "une"
...
+                n -> shiftL (fromIntegral ((n - 15) + 127)) 23
+            f = sign .|. expo .|. signif
+            f' = castWord32ToFloat f
+        in
+        float f'
+  ValBFloat (FPBFloat x) ->
+    -- BFloat half floats have a 1-bit sign, 8-bit exponent, and 7-bit
+    -- significand.
+    let expbits = x .&. 0x7f80 in
+    if expbits == 0x7f80 -- isInfinite || isNan
```
RyanGlScott: Similarly to what is used above:


```suggestion
    if expbits == 0x7f80 -- true if x is an infinite or NaN value
```

RyanGlScott (approved):
It would be worth referencing https://github.com/GaloisInc/llvm-pretty/issues/66 in the commit messages of the commits that are related to it. It might also be worth referencing https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/214 in the commit that adds `BFloat`.

.github/workflows/haskell.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v1
```
langston-barrett: @sauclovian-g I see you've been hitting some cache-related problems. That's likely due to https://github.com/GaloisInc/.github/issues/65, which is fixed in more recent versions. We could try bumping this to `v2` if you're game.

```suggestion
    uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v2
```

src/Text/LLVM/AST.hs
```diff
@@ -713,10 +718,13 @@ primTypeNull (FloatType ft) = floatTypeNull ft
...
 floatTypeNull :: FloatType -> Value' lab
+floatTypeNull Half     = ValHalf $ FPHalf 0
+floatTypeNull BFloat   = ValBFloat $ FPBFloat 0
 floatTypeNull Float    = ValFloat 0
-floatTypeNull Double   = ValDouble 0 -- XXX not sure about this
+floatTypeNull Double   = ValDouble 0
+floatTypeNull Fp128    = ValFP128 $ FP128_LongDouble 0 0
 floatTypeNull X86_fp80 = ValFP80 $ FP80_LongDouble 0 0
-floatTypeNull _        = error "must be a float type"
+floatTypeNull PPC_fp128 = ValFP128_PPC $ FP128_PPC_DoubleDouble 0 0
```
langston-barrett, line 727: Nice 😄

src/Text/LLVM/AST.hs
```diff
@@ -1560,9 +1572,32 @@ data Value' lab
 
 type Value = Value' BlockLabel
 
+-- | 16-bit half-precision floating point value (IEEE half)
+data FPHalfValue = FPHalf Word16
+               deriving (Data, Eq, Ord, Generic, Show)
```
langston-barrett: Minor point, but why are these `deriving`s way over to the right? Other types have them at indentation 4, which seems reasonable.

src/Text/LLVM/PP.hs
```diff
@@ -865,11 +868,52 @@ ppFCmpOp Fune   = "une"
 ppFCmpOp Funo   = "uno"
 ppFCmpOp Ftrue  = "true"
 
+-- | Pad a string by prepending '0'.
+zerofill :: Int -> String -> String
+zerofill width s =
```
langston-barrett, line 873: nit: use `camelCase`

```suggestion
zeroFill :: Int -> String -> String
zeroFill width s =
```

langston-barrett: Similar comment applies elsewhere, e.g., `expbits`

src/Text/LLVM/PP.hs
```diff
@@ -865,11 +868,52 @@ ppFCmpOp Fune   = "une"
...
 ppValue' pp val = case val of
   ValInteger i       -> integer i
   ValBool b          -> ppBool b
   -- Note: for +Inf/-Inf/NaNs, we want to output the bit-correct sequence
+  ValHalf (FPHalf x) ->
+    -- Half floats have a 1-bit sign, 5-bit exponent, and 10-bit
+    -- significand.
+    let expbits = x .&. 0x7c00 in
+    if expbits == 0x7c00 -- true if x is an infinite or NaN value
+      then text "0xH" <> text (zerofill 4 $ showHex x "")
```
langston-barrett, line 936: What does the "H" represent here? ("hex" comes to mind, but 0x already indicates hex?)

src/Text/LLVM/PP.hs
```diff
@@ -865,11 +868,52 @@ ppFCmpOp Fune   = "une"
...
+        let -- Convert to 32-bit single float (1-bit sign, 8-bit exponent,
+            -- 23-bit significand)
+            sign = shiftL (fromIntegral (x .&. 0x8000)) (31 - 15)
+            signif = shiftL (fromIntegral (x .&. 0xfff)) (23 - 10)
+            expo = case fromIntegral @Word16 @Int (shiftR expbits 10) of
+                0 -> 0
+                n -> shiftL (fromIntegral ((n - 15) + 127)) 23
+            f = sign .|. expo .|. signif
+            f' = castWord32ToFloat f
+        in
```
langston-barrett: Yet another minor stylistic comment: for blocks like this that perform a single piece of readily-understood, context-independent computation, I like to put them in a `where` or a whole separate function. Having a name like `halfToFloat` (and a matching type signature) would clarify this code, IMO. (Though the comment does already make the intent clear, thanks for that!)

langston-barrett: Similar comment applies elsewhere

src/Text/LLVM/PP.hs
langston-barrett: Oh, based on the below, perhaps this is also a LLVM LangRef choice? Could we get a comment/URL here as well?

langston-barrett (approved):
Mostly minor stylistic suggestions, feel free to take them or leave them!

src/Text/LLVM/AST.hs
sauclovian-g: No idea, I just copied the FP80 one. Fixed

src/Text/LLVM/PP.hs
sauclovian-g: I fixed `expBits`. "zerofill" is one word :-p

src/Text/LLVM/PP.hs
sauclovian-g: Yes, fixed.

src/Text/LLVM/PP.hs
sauclovian-g: I've moved a bunch of this stuff out of line.

sauclovian-g:
For the record, the force-push retconned the commit messages as Ryan suggested, and I also took the opportunity to squash most of the previous commits, as there wasn't any real reason to keep them separate.

PR #188, opened 2026-01-31
Fix how single-precision floats get printed in hex

### LINKED ISSUE #187: We don't print single-precision floats in hex correctly
Adding tests for floating point constants to llvm-pretty-bc-parser reveals that when we print single-precision floats as hex, we don't do it correctly.

What we do: convert the `Float` value to a 32-bit integer and print it as hex.

What we're supposed to do: convert the `Float` value to a `Double`, then convert it to a 64-bit integer and print that as hex.

sauclovian-g:
They're supposed to be promoted to double and then printed as 64-bit integers.

Closes #187.

### DIFF
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -28,7 +28,7 @@ import Data.Char (isAlphaNum,isAscii,isDigit,isPrint,ord,toUpper)
 import Data.List ( intersperse, nub )
 import qualified Data.Map as Map
 import Data.Maybe (catMaybes,fromMaybe,isJust)
-import GHC.Float (castDoubleToWord64, castFloatToWord32, castWord32ToFloat, float2Double)
+import GHC.Float (castDoubleToWord64, castWord32ToFloat, float2Double)
 import Numeric (showHex)
 import Text.PrettyPrint.HughesPJ
 import Data.Int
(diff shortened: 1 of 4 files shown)


sauclovian-g:
Ok, that should _not_ have passed. 👀

sauclovian-g:
Yeah, it didn't actually recompile. I think we have not fixed the cache problems 😿

sauclovian-g:
and this time it recompiled the test program but not the library...

src/Text/LLVM/PP.hs
```diff
@@ -943,7 +943,7 @@ ppValue' pp val = case val of
       else float $ castWord32ToFloat $ bfloatToSingleBits x
   ValFloat f         ->
     if isInfinite f || isNaN f
-      then text "0x" <> text (showHex (castFloatToWord32 f) "")
+      then text "0x" <> text (showHex (castDoubleToWord64 $ float2Double f) "")
```
RyanGlScott, line 948: For symmetry with the other floating-point cases, it would be worth citing the LLVM Language Reference as justification for extending to a double here, i.e.,

```suggestion
      -- shown as 0x<<16-hex-digits>>, per
      -- https://llvm.org/docs/LangRef.html#simple-constants
      then text "0x" <> text (showHex (castDoubleToWord64 $ float2Double f) "")
```

src/Text/LLVM/PP.hs
RyanGlScott, line 954: While we're at it, we may as well add a similar comment here:


```suggestion
      -- shown as 0x<<16-hex-digits>>, per
      -- https://llvm.org/docs/LangRef.html#simple-constants
      then text "0x" <> text (showHex (castDoubleToWord64 d) "")
```

RyanGlScott (approved):
Good catch!

Can you mention this bugfix in the README?

kquick (approved):

PR #189, opened 2026-02-25
Support metadata fields added in LLVM 22

RyanGlScott:
This adds `dicuSourceLanguageVersion :: Word64` to `DICompileUnit'` and `dibtDataSize :: Word32` to `DIBasicType'` to reflect similar changes in LLVM 22. Together, these pave the way for LLVM 22 bitcode parsing support in `llvm-pretty-bc-parser` (see https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/339 and https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/340, respectively).

### DIFF
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -1342,6 +1342,12 @@ ppDICompileUnit' pp cu = "!DICompileUnit"
        ,     (("sdk:"                   <+>) . doubleQuotes . text)
              <$> (dicuSDK cu)
        ]
+       ++
+       when' (llvmVer >= 22)
+       [ if dicuSourceLanguageVersion cu > 0
+         then pure ("sourceLanguageVersion:" <+> integral (dicuSourceLanguageVersion cu))
+         else Nothing
+       ]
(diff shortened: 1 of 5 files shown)


kquick (approved):

langston-barrett (approved):
Should we also update the README? It says we support up to LLVM 17.

src/Text/LLVM/PP.hs
```diff
@@ -75,7 +75,7 @@ llvmV3_8 = 3
 -- this is used for defaulting and otherwise reporting the maximum LLVM version
 -- known to be supported.
 llvmVlatest :: LLVMVer
-llvmVlatest = 19
+llvmVlatest = 22
```
langston-barrett: Perhaps we could add a comment here saying that when this is updated the README should be too (and one in the README pointing here).

RyanGlScott: Good call. I've updated the README to mention LLVM 22 support, and I've also left reminders both in the README and in the code here to update the other.

PR #190, opened 2026-03-06
Add documentation for Linkage constructors.

kquick:


src/Text/LLVM/AST.hs
```diff
@@ -918,22 +918,47 @@ brTargets (BasicBlock _ stmts) =
 
 -- Attributes ------------------------------------------------------------------
 
--- | Symbol Linkage
+-- | Symbol Linkage provides information on how the symbol should be handled
```
langston-barrett:
```suggestion
-- | Symbol linkage provides information on how the symbol should be handled
```
or `'Linkage'` if a reference to the type is meant

src/Text/LLVM/AST.hs
```diff
@@ -918,22 +918,47 @@ brTargets (BasicBlock _ stmts) =
...
+    -- linking to avoid collisions. Not visible in the object file's symbol table.
   | LinkerPrivate
   | LinkerPrivateWeak
   | LinkerPrivateWeakDefAuto
   | Internal
+    -- ^ Similar to private but shows up as a local symbol (e.g. C @static@)
   | AvailableExternally
+    -- ^ External declaration, never defined in the object file. Allows inlining
+    -- and other optimizations knowing the symbol exists externally.  May be
+    -- discarded at will.  Only allowed on Declare, not on Define.
```
langston-barrett:
```suggestion
    -- discarded at will.  Only allowed on 'Declare', not on 'Define'.
```

langston-barrett (approved):
A few other places need Haddock links too but generally LGTM. Thanks!

PR #191, opened 2026-04-03
Add atFileLines to DebugUtils and use in PP for new ppModuleAtLines.

kquick:
The ppModuleAtLines allows printing a subset of a function (Define) or a global definition that matches the specified source file and line.

### DIFF
--- a/src/Text/LLVM/DebugUtils.hs
+++ b/src/Text/LLVM/DebugUtils.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# Language TransformListComp, MonadComprehensions #-}
 {- |
 Module           : Text.LLVM.DebugUtils
@@ -32,18 +34,25 @@ module Text.LLVM.DebugUtils
   -- * Line numbers of definitions
   , debugInfoGlobalLines
   , debugInfoDefineLines
+  , atFileLines
+  , AtFileLines(atDefine, atBlockStart, atStmt, atGlobal)
+  , DefineRel(..)
+  , BlockRel(..)
   ) where
 
 import           Control.Applicative    ((<|>))
 import           Control.Monad          ((<=<))
 import           Data.Bits              (Bits(..))
+import           Data.Bool              (bool)
 import           Data.IntMap            (IntMap)
 import qualified Data.IntMap as IntMap
 import           Data.List              (elemIndex, tails, stripPrefix)
 import           Data.Map               (Map)
 import qualified Data.Map    as Map
-import           Data.Maybe             (fromMaybe, listToMaybe, maybeToList, mapMaybe)
+import           Data.Maybe (fromMaybe, listToMaybe, maybeToList, mapMaybe)
 import           Data.Word              (Word16, Word64)
+import           Lens.Micro.Platform    ((^.), at, _Just, to)
+import           System.FilePath        ( (</>), equalFilePath, normalise )
 import           Text.LLVM.AST
 
 dbgKind :: String
@@ -170,6 +179,31 @@ getDebugInfo :: MdMap -> ValMd -> Maybe DebugInfo
 getDebugInfo mdMap (ValMdRef i)    = getDebugInfo mdMap =<< IntMap.lookup i mdMap
 getDebugInfo _ (ValMdDebugInfo di) = Just di
 getDebugInfo _ _                   = Nothing
+
+getMDFile :: MdMap -> ValMd -> Maybe FilePath
+getMDFile mdMap = \case
+  ValMdLoc l -> getMDFile mdMap $ dlScope l
+  ValMdRef i -> mdMap ^. at i . _Just . to (getMDFile mdMap)
+  ValMdDebugInfo (DebugInfoGlobalVariable gv) ->
+    (getMDFile mdMap =<< digvFile gv) <|> (getMDFile mdMap =<< digvScope gv)
+  ValMdDebugInfo (DebugInfoLocalVariable lv) ->
+    (getMDFile mdMap =<< dilvFile lv) <|> (getMDFile mdMap =<< dilvScope lv)
+  ValMdDebugInfo (DebugInfoSubprogram sp) ->
+    (getMDFile mdMap =<< dispFile sp) <|> (getMDFile mdMap =<< dispScope sp)
+  ValMdDebugInfo (DebugInfoLexicalBlock lb) ->
+    (getMDFile mdMap =<< dilbFile lb) <|> (getMDFile mdMap =<< dilbScope lb)
+  ValMdDebugInfo (DebugInfoLexicalBlockFile lf) ->
+    (getMDFile mdMap =<< dilbfFile lf) <|> (getMDFile mdMap $ dilbfScope lf)
+  ValMdDebugInfo (DebugInfoDerivedType dt) ->
+    (getMDFile mdMap =<< didtFile dt) <|> (getMDFile mdMap =<< didtScope dt)
+  ValMdDebugInfo (DebugInfoCompositeType ct) ->
+    (getMDFile mdMap =<< dictFile ct) <|> (getMDFile mdMap =<< dictScope ct)
+  ValMdDebugInfo (DebugInfoCompileUnit cu) -> getMDFile mdMap =<< dicuFile cu
+  ValMdDebugInfo (DebugInfoNameSpace ns) -> getMDFile mdMap $ dinsFile ns
+  ValMdDebugInfo (DebugInfoLabel bl) ->
+    (getMDFile mdMap =<< dilFile bl) <|> (getMDFile mdMap =<< dilScope bl)
+  ValMdDebugInfo (DebugInfoFile i) -> pure $ difDirectory i </> difFilename i
+  _ -> Nothing
 
 getInteger :: MdMap -> ValMd -> Maybe Integer
 getInteger mdMap (ValMdRef i)                          = getInteger mdMap =<< IntMap.lookup i mdMap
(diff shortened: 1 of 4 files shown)


src/Text/LLVM/DebugUtils.hs
```diff
@@ -171,6 +180,31 @@ getDebugInfo mdMap (ValMdRef i)    = getDebugInfo mdMap =<< IntMap.lookup i mdMa
...
+    (getMDFile mdMap =<< dilbfFile lf) <|> (getMDFile mdMap $ dilbfScope lf)
+  ValMdDebugInfo (DebugInfoDerivedType dt) ->
+    (getMDFile mdMap =<< didtFile dt) <|> (getMDFile mdMap =<< didtScope dt)
+  ValMdDebugInfo (DebugInfoCompositeType ct) ->
+    (getMDFile mdMap =<< dictFile ct) <|> (getMDFile mdMap =<< dictScope ct)
+  ValMdDebugInfo (DebugInfoCompileUnit cu) -> getMDFile mdMap =<< dicuFile cu
+  ValMdDebugInfo (DebugInfoNameSpace ns) -> getMDFile mdMap $ dinsFile ns
+  ValMdDebugInfo (DebugInfoLabel bl) ->
+    (getMDFile mdMap =<< dilFile bl) <|> (getMDFile mdMap =<< dilScope bl)
+  ValMdDebugInfo (DebugInfoFile i) -> pure $ difDirectory i </> difFilename i
```
RyanGlScott: Unless I've overlooked something, this is the only case that doesn't recursively call `getMDFile`. I think it would be worth putting this case at the top to make it more obvious that the recursion stops here.

src/Text/LLVM/PP.hs
```diff
@@ -1581,6 +1587,46 @@ ppDIArgList' pp args = "!DIArgList"
...
+-- situations where additional or alternative pretty-printing functionality is
+-- needed.
+
+ppModuleAtLines :: (?config :: Config) => String -> Integer -> Fmt Module
+ppModuleAtLines file line =
+  toDoc . atFileLines (AddDocAtLine ?config) (Start empty) file line
+
+data AddDocAtLine = AddDocAtLine Config
+
+data DocBld = Start Doc | DF DocBld Define Doc | BS DocBld (Maybe BlockLabel) Doc
```
RyanGlScott: Please add some Haddocks for this type. What do `Start`, `DF`, and `BS` represent?

src/Text/LLVM/PP.hs
```diff
@@ -1581,6 +1587,46 @@ ppDIArgList' pp args = "!DIArgList"
...
+-- Auxiliary pretty-printing functions
+--
+-- These are alternative pretty-printing functions (intead of the pretty-printers
+-- for the basic AST elements above). These functions can be used in
+-- situations where additional or alternative pretty-printing functionality is
+-- needed.
+
+ppModuleAtLines :: (?config :: Config) => String -> Integer -> Fmt Module
+ppModuleAtLines file line =
+  toDoc . atFileLines (AddDocAtLine ?config) (Start empty) file line
```
RyanGlScott: Is this the only function that's meant to be part of the public API, and are all the functions added below it meant to be internal? If so, it might be worth adding an explicit export list to `Text.LLVM.PP` to enforce this.

src/Text/LLVM/PP.hs
```diff
@@ -1581,6 +1587,46 @@ ppDIArgList' pp args = "!DIArgList"
...
+
+-- -------------------------------------------------------------------
+-- Auxiliary pretty-printing functions
+--
+-- These are alternative pretty-printing functions (intead of the pretty-printers
+-- for the basic AST elements above). These functions can be used in
+-- situations where additional or alternative pretty-printing functionality is
+-- needed.
+
+ppModuleAtLines :: (?config :: Config) => String -> Integer -> Fmt Module
```
RyanGlScott: The name of this function is `ppModuleAtLines`, but it appears that the function only takes a single `line :: Integer` argument. Am I missing something?

Either way, I think this function deserves some Haddocks to point out how it differs from `ppModule`. An example would also be helpful to illustrate this further.

src/Text/LLVM/DebugUtils.hs
```diff
@@ -514,3 +548,96 @@ debugInfoDefineLines = Map.fromList . mapMaybe go . modUnnamedMd
...
+             , dr == FirstBlock True && null (stmtMetadata s)
+             ]
+  in foldl onGlobal (foldl onDefs seed $ modDefines mdule) $ modGlobals mdule
+
+-- | The handler passed to 'atFileLines' must be an instance of this class.
+--
+-- Here, @b@ is an object for which the following methods can be called with an
+-- accumulator @a@ and the corresponding LLVM AST element that has the @lab@
+-- label type.  The method will return an updated accumulator.
+class AtFileLines a b where
```
RyanGlScott, line 618: I have to admit that this seems a bit over-engineered, especially since there is only a single instance of this type class. Do you foresee ever needing to be this polymorphic? If not, I would suggest just monomorphizing this to `DocBld` and `AddDocAtLine`.

src/Text/LLVM/DebugUtils.hs
```diff
@@ -514,3 +548,96 @@ debugInfoDefineLines = Map.fromList . mapMaybe go . modUnnamedMd
...
+atFileLines handle seed file line mdule =
+  let mdMap = mkMdMap mdule
+      matchesFile x =
+        let fn = normalise file
+            fl = length fn
+            xl = length x
+            (p,r) = splitAt (xl - fl) x
+        in and [ fl <= xl
+               , fn `equalFilePath` r
+               , null p || "/" == (take 1 $ reverse p)
```
RyanGlScott: Style nit pick: same number of parentheses, but fewer infix operators:


```suggestion
               , null p || "/" == take 1 (reverse p)
```

src/Text/LLVM/DebugUtils.hs
```diff
@@ -32,18 +34,25 @@ module Text.LLVM.DebugUtils
...
+import           Data.Bool              (bool)
 import           Data.IntMap            (IntMap)
 import qualified Data.IntMap as IntMap
 import           Data.List              (elemIndex, tails, stripPrefix)
 import           Data.Map               (Map)
 import qualified Data.Map    as Map
-import           Data.Maybe             (fromMaybe, listToMaybe, maybeToList, mapMaybe)
+import           Data.Maybe (fromMaybe, listToMaybe, maybeToList, mapMaybe)
 import           Data.Word              (Word16, Word64)
+import           Lens.Micro.Platform    ((^.), at, _Just, to)
```
RyanGlScott, line 54: Do we really need to incur an extra `microlens-platform` dependency here? As far as I can tell, all of these functions are available in just `microlens` (which we already depend on).

src/Text/LLVM/DebugUtils.hs
```diff
@@ -32,18 +34,25 @@ module Text.LLVM.DebugUtils
...
 import           Control.Monad          ((<=<))
 import           Data.Bits              (Bits(..))
+import           Data.Bool              (bool)
 import           Data.IntMap            (IntMap)
 import qualified Data.IntMap as IntMap
 import           Data.List              (elemIndex, tails, stripPrefix)
 import           Data.Map               (Map)
 import qualified Data.Map    as Map
-import           Data.Maybe             (fromMaybe, listToMaybe, maybeToList, mapMaybe)
+import           Data.Maybe (fromMaybe, listToMaybe, maybeToList, mapMaybe)
```
RyanGlScott: Any reason for changing the whitespace here to be different from the rest of the imports?

src/Text/LLVM/DebugUtils.hs
kquick: `microlens` has no instance of `IxValue` for `IntMap`... that comes from `microlens-platform`.

src/Text/LLVM/DebugUtils.hs
kquick: There are two bits of functionality being added by this PR:
1. The ability to process an LLVM module for the portion of the module that corresponds to a source location
2. The ability to pretty-print a portion of an LLVM module based on specifying a source location

Where number 2 uses number 1 but number 1 does not rely on pretty printing at all and is anticipated to have other uses.

src/Text/LLVM/DebugUtils.hs
kquick: Done.

src/Text/LLVM/PP.hs
kquick: This is really only meant to be used internally by the corresponding methods above.  Per the note below, these are now private to the module and thus as internals, I added some brief comments but not full haddocks.

src/Text/LLVM/PP.hs
kquick: Yes, it is intended as the only export.  

I initially tried moving it to a submodule of `Text.LLVM.PP` so that only this function could be exported by that submodule (and thus by `Text.LLVM.PP`, but this turned out to need 3 submodules (one for the internal definitions in `Text.LLVM.PP`, one for most of `Text.LLVM.PP`'s exports, and one for this new functionality) which was a pretty invasive set of changes, so I opted instead for the explicit export list.  I did leave out a few other things that looked like they really should only be intended for internal use (e.g. `commas`, `angles`, `structBraces`, `Fmt`,  `opt`, `integral`, and a few others) but since this module did not previously have controlled exports, we may find we need to expand the export list for existing downstream uses.

src/Text/LLVM/PP.hs
kquick: Good catch.   Renamed to `ppModuleAtLine`.  I also added some Haddocks.

src/Text/LLVM/DebugUtils.hs
kquick: I tend toward the original form because it's clearer when the entire RHS is in parentheses (slightly less brainwattage needed to parse), but it's not really a big issue for this instance: changed.

src/Text/LLVM/DebugUtils.hs
kquick: I got tired of looking at the >80 char line during dev.  Reverted.

src/Text/LLVM/DebugUtils.hs
```diff
@@ -514,3 +549,96 @@ debugInfoDefineLines = Map.fromList . mapMaybe go . modUnnamedMd
...
+atFileLines handle seed file line mdule =
+  let mdMap = mkMdMap mdule
+      matchesFile x =
+        let fn = normalise file
+            fl = length fn
+            xl = length x
+            (p,r) = splitAt (xl - fl) x
+        in and [ fl <= xl
+               , fn `equalFilePath` r
+               , null p || "/" == take 1 (reverse p)
```
RyanGlScott: A bit clearer:


```suggestion
               , null p || hasTrailingPathSeparator p
```

src/Text/LLVM/PP.hs
```diff
@@ -1581,6 +1732,66 @@ ppDIArgList' pp args = "!DIArgList"
...
+-- situations where additional or alternative pretty-printing functionality is
+-- needed.
+
+-- | This is an auxiliary pretty printer for showing just part of a module: the
+-- part corresponding to a specific source file and line in the source file
+-- (whereas ppModule or even ppDefine will show the *entire* module or
+-- definition/function).  A range of lines can be displayed by iterative calls
+-- over multiple lines.
+--
+-- > putStrLn$ ppLLVM llvmVlatest $ ppModuleAtLine "foo.c" 23 llvmModule
```
RyanGlScott:
```suggestion
-- > putStrLn $ ppLLVM llvmVlatest $ ppModuleAtLine "foo.c" 23 llvmModule
```

src/Text/LLVM/PP.hs
```diff
@@ -1581,6 +1732,66 @@ ppDIArgList' pp args = "!DIArgList"
 ppDIArgList :: Fmt DIArgList
 ppDIArgList = ppDIArgList' ppLabel
 
+
+-- -------------------------------------------------------------------
+-- Auxiliary pretty-printing functions
+--
+-- These are alternative pretty-printing functions (intead of the pretty-printers
```
RyanGlScott:
```suggestion
-- These are alternative pretty-printing functions (instead of the pretty-printers
```

src/Text/LLVM/DebugUtils.hs
RyanGlScott: Ah, OK! Perhaps it would be worth a comment here saying that this import is also used to bring in the orphan `IxValue` instance for `IntMap`---that part wasn't obvious at a first glance.

src/Text/LLVM/DebugUtils.hs
RyanGlScott: OK. Would you mind documenting this new functionality in the changelog? If it's meant to have other downstream uses, we should make sure that we advertise its existence.

src/Text/LLVM/DebugUtils.hs
kquick: TIL!

RyanGlScott (approved):

PR #192, opened 2026-04-07
Use UnnamedMdIdx newtype wrapper for Unnamed metadata indices.

kquick:
This makes them distinct and explicit.

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -287,18 +287,18 @@ data NamedMd = NamedMd
 
 -- Unnamed Metadata ------------------------------------------------------------
 
--- | This is the type used to represent an Unnamed Metadata index.  A newtype
+-- | This is the type used to represent an unnamed metadata index.  A newtype
 -- wrapper is used to distinguish the specific use of this value as this
 -- particular type of index.
 --
 -- The Ord instance is provided to allow these indices to be used as Map keys.
--- Although Num and Enum are provided to enable manipulation, care should be
--- taken that these are all very carefully used only where needed and
--- appropriate.  In general, the `nextUnnamedMdIdx` is preferred.
+-- Although Num and Enum instances are provided to enable manipulation, care
+-- should be taken that these are all very carefully used only where needed and
+-- appropriate.  In general, the `nextUnnamedMdIdx` function is preferred.
 newtype UnnamedMdIdx = UnnamedMdIdx { unnamedMdIdx :: Int }
   deriving (Data, Eq, Generic, Ord, Enum, Num, Show)
 
--- | This is used when constructing an AST and a new Unnamed Metadata element is
+-- | This is used when constructing an AST and a new unnamed metadata element is
 -- to be added.  It should be passed the current maximum known index and will
 -- return the new, unused index that should be used.
 nextUnnamedMdIdx :: UnnamedMdIdx -> UnnamedMdIdx
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -287,9 +287,20 @@ data NamedMd = NamedMd
 
 -- Unnamed Metadata ------------------------------------------------------------
 
+-- | This is the type used to represent an Unnamed Metadata index.  A newtype
+-- wrapper is used to distinguish the specific use of this value as this
+-- particular type of index.
+--
+-- The Ord instance is provided to allow these indices to be used as Map keys.
+-- Although Num and Enum are provided to enable manipulation, care should be
+-- taken that these are all very carefully used only where needed and
+-- appropriate.  In general, the `nextUnnamedMdIdx` is preferred.
 newtype UnnamedMdIdx = UnnamedMdIdx { unnamedMdIdx :: Int }
   deriving (Data, Eq, Generic, Ord, Enum, Num, Show)
 
+-- | This is used when constructing an AST and a new Unnamed Metadata element is
+-- to be added.  It should be passed the current maximum known index and will
+-- return the new, unused index that should be used.
 nextUnnamedMdIdx :: UnnamedMdIdx -> UnnamedMdIdx
 nextUnnamedMdIdx (UnnamedMdIdx i) = UnnamedMdIdx $ i + 1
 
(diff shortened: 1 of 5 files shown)


src/Text/LLVM/AST.hs
```diff
@@ -278,14 +281,31 @@ emptyModule  = Module
...
 data NamedMd = NamedMd
   { nmName   :: String
-  , nmValues :: [Int]
+  , nmValues :: [UnnamedMdIdx]
   } deriving (Data, Eq, Generic, Ord, Show)
 
 
 -- Unnamed Metadata ------------------------------------------------------------
 
+newtype UnnamedMdIdx = UnnamedMdIdx { unnamedMdIdx :: Int }
```
RyanGlScott, line 298: Please add some Haddocks for this type.

src/Text/LLVM/AST.hs
```diff
@@ -278,14 +281,31 @@ emptyModule  = Module
...
+  deriving (Data, Eq, Generic, Ord, Enum, Num, Show)
+
+nextUnnamedMdIdx :: UnnamedMdIdx -> UnnamedMdIdx
```
RyanGlScott, line 304: "Next" in what sense? Does this assume that you are always passing it the value of the largest `UnnamedMdIdx` encountered thus far? Some documentation would be helpful here.

src/Text/LLVM/AST.hs
```diff
@@ -278,14 +281,31 @@ emptyModule  = Module
...
   { nmName   :: String
-  , nmValues :: [Int]
+  , nmValues :: [UnnamedMdIdx]
   } deriving (Data, Eq, Generic, Ord, Show)
 
 
 -- Unnamed Metadata ------------------------------------------------------------
 
+newtype UnnamedMdIdx = UnnamedMdIdx { unnamedMdIdx :: Int }
+  deriving (Data, Eq, Generic, Ord, Enum, Num, Show)
```
RyanGlScott, line 299: I can see why it would be convenient to offer `Enum` and `Num` instances for this type, although there is a risk that a user can misuse these instances to construct invalid `UnnamedMdIdx` values (e.g., by writing `idx + 50` or some such). I think it would be worth adding a disclaimer that, to the extent that is possible, users should only construct `UnnamedMdIdx` values by using the `nextUnnamedMdIdx` and `nonNullUnnamedMdIdx` functions.

src/Text/LLVM/AST.hs
```diff
@@ -278,14 +281,31 @@ emptyModule  = Module
...
+-- | In the bitcode, an "optional" unnamed metadata index is indicated by 0 (not
+-- present) or the actual index + 1.  The parsing should treat these optional as
+-- a different type than an UnnamedMdIdx, but this is not presently detected at
+-- the parsing level, so this function is used to convert a parsed UnnamedMdIdx
+-- to the option of the correct index.
+--
+-- Note that it is NOT valid to call this twice (or not at all in the event you
+-- are starting with an optional form).
+nonNullUnnamedMdIdx :: UnnamedMdIdx -> UnnamedMdIdx
+nonNullUnnamedMdIdx (UnnamedMdIdx i) = UnnamedMdIdx $ i - 1
```
RyanGlScott: See my comments about the type of this function in https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/352#discussion_r3045166189. I am very skeptical about always computing `i - 1`, even when the argument is zero (i.e., optional).

kquick: Good call.  I wasn't sure of the scope of these and took a minimalist approach to the change, but this is a better approach and doesn't seem to require too many updates.  Ultimately, I'd prefer a different source type (`OptionalUnnamedMdIdx` or something like that) but I'm not sure if the Parser knows when to make which one, and so I figured I'd defer that for a future set of changes once these land.

src/Text/LLVM/AST.hs
```diff
@@ -278,14 +281,43 @@ emptyModule  = Module
...
 
 
 -- Unnamed Metadata ------------------------------------------------------------
 
+-- | This is the type used to represent an Unnamed Metadata index.  A newtype
+-- wrapper is used to distinguish the specific use of this value as this
+-- particular type of index.
+--
+-- The Ord instance is provided to allow these indices to be used as Map keys.
+-- Although Num and Enum are provided to enable manipulation, care should be
```
RyanGlScott:
```suggestion
-- Although Num and Enum instances are provided to enable manipulation, care should be
```

src/Text/LLVM/AST.hs
```diff
@@ -278,14 +281,43 @@ emptyModule  = Module
...
+-- taken that these are all very carefully used only where needed and
+-- appropriate.  In general, the `nextUnnamedMdIdx` is preferred.
```
RyanGlScott:
```suggestion
-- appropriate.  In general, the `nextUnnamedMdIdx` function is preferred.
```

src/Text/LLVM/AST.hs
```diff
@@ -278,14 +281,43 @@ emptyModule  = Module
...
 data NamedMd = NamedMd
   { nmName   :: String
-  , nmValues :: [Int]
+  , nmValues :: [UnnamedMdIdx]
   } deriving (Data, Eq, Generic, Ord, Show)
 
 
 -- Unnamed Metadata ------------------------------------------------------------
 
+-- | This is the type used to represent an Unnamed Metadata index.  A newtype
```
RyanGlScott: I'm not sure that this is a proper noun:


```suggestion
-- | This is the type used to represent an unnamed metadata index.  A newtype
```

src/Text/LLVM/AST.hs
```diff
@@ -278,14 +281,43 @@ emptyModule  = Module
...
+-- particular type of index.
+--
+-- The Ord instance is provided to allow these indices to be used as Map keys.
+-- Although Num and Enum are provided to enable manipulation, care should be
+-- taken that these are all very carefully used only where needed and
+-- appropriate.  In general, the `nextUnnamedMdIdx` is preferred.
+newtype UnnamedMdIdx = UnnamedMdIdx { unnamedMdIdx :: Int }
+  deriving (Data, Eq, Generic, Ord, Enum, Num, Show)
+
+-- | This is used when constructing an AST and a new Unnamed Metadata element is
```
RyanGlScott: Similarly:


```suggestion
-- | This is used when constructing an AST and a new unnamed metadata element is
```

RyanGlScott (approved):

PR #193, opened 2026-04-13
Export Fmt from PP and add class for pretty printing lab parameter

kquick:
The recently added explicit export controls for the LLVM pretty printer excluded the `Fmt` definition which is needed externally by some projects.

This also adds a `LabelPrinter` class which allows client code to keep the `lab` type parameter as a type variable yet still invoke the pretty printing.

### DIFF
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -57,7 +57,7 @@ module Text.LLVM.PP
   , ppFunAttr
   , ppLabelDef
   , ppLabel
-  , LabelPrinter(labelPrinter)
+  , PrettyLabel(ppLabel')
   , ppBasicBlock
   , ppStmt
   , ppAttachedMetadata
@@ -629,31 +629,32 @@ ppBasicBlock bb = ppMaybe ppLabelDef (bbLabel bb)
               $+$ nest 2 (vcat (map ppStmt (bbStmts bb)))
 
 
--- | Most of the pretty printing functions in this module are written for the
--- parameterized label data form and thus take a Fmt argument for handling the
--- parameterized label type.
+-- | Most of the pretty printing functions (based on 'Pretty') in this module are
+-- written for the parameterized label data form and thus take a 'Fmt' argument
+-- for handling the parameterized label type.
 --
--- When the parameterized label type is known to be 'BlockLabel', then ppLabel
+-- When the parameterized label type is known to be 'BlockLabel', then 'ppLabel'
 -- can be passed as this first argument to those pretty printing functions.
 --
 -- When the parameterized label type is not known (e.g. when implementing other
--- libraries that utilize llvm-pretty), the library may need to invoke the pretty
--- printer without passing an explicit label printer, and can instead use the
--- following class as a constraint for an instance that will be resolved later.
+-- libraries that utilize @llvm-pretty@), the library may need to invoke the
+-- pretty printer without passing an explicit label printer, and can instead use
+-- the following class as a constraint for an instance that will be resolved
+-- later.
 --
 -- For example:
 --
 -- > data Something lab = Something { ..., val :: Value' lab, ... }
 -- >
--- > instance LabelPrinter lab => Pretty Something where
--- >   pretty s = .... <> ppValue' astLabelPrinter <>
+-- > instance PrettyLabel lab => Pretty Something where
+-- >   pretty s = .... <> ppValue' ppLabel' <>
 --
 
-class LabelPrinter lab where
-  labelPrinter :: Fmt lab
+class PrettyLabel lab where
+  ppLabel' :: Fmt lab
 
-instance LabelPrinter BlockLabel where
-  labelPrinter = ppLabel
+instance PrettyLabel BlockLabel where
+  ppLabel' = ppLabel
 
 
 -- Statements ------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/Text/LLVM/PP.hs
```diff
@@ -627,6 +629,33 @@ ppBasicBlock bb = ppMaybe ppLabelDef (bbLabel bb)
...
+-- For example:
+--
+-- > data Something lab = Something { ..., val :: Value' lab, ... }
+-- >
+-- > instance LabelPrinter lab => Pretty Something where
+-- >   pretty s = .... <> ppValue' astLabelPrinter <>
+--
+
+class LabelPrinter lab where
+  labelPrinter :: Fmt lab
```
RyanGlScott: I think the Haddocks should reference the [`Pretty`](https://hackage.haskell.org/package/pretty-1.1.3.6/docs/Text-PrettyPrint-Annotated-HughesPJClass.html#t:Pretty) class somewhere. `LabelPrinter` is very nearly the same class, except that this class constrains its method with a `(?config :: Config`) constraint.

src/Text/LLVM/PP.hs
```diff
@@ -627,6 +629,33 @@ ppBasicBlock bb = ppMaybe ppLabelDef (bbLabel bb)
...
+-- libraries that utilize llvm-pretty), the library may need to invoke the pretty
+-- printer without passing an explicit label printer, and can instead use the
+-- following class as a constraint for an instance that will be resolved later.
+--
+-- For example:
+--
+-- > data Something lab = Something { ..., val :: Value' lab, ... }
+-- >
+-- > instance LabelPrinter lab => Pretty Something where
+-- >   pretty s = .... <> ppValue' astLabelPrinter <>
```
RyanGlScott: What is `astLabelPrinter`?

src/Text/LLVM/PP.hs
RyanGlScott: The naming convention used here deviates somewhat from the rest of this module: nearly all other pretty-printing functions start with `pp*`, but `labelPrinter` does not. Should it?

Relatedly, I wonder if this class should be called something like `PrettyLabel` to more clearly emphasize that it's used for pretty-printing.

src/Text/LLVM/PP.hs
```diff
@@ -627,6 +629,33 @@ ppBasicBlock bb = ppMaybe ppLabelDef (bbLabel bb)
...
 
+-- | Most of the pretty printing functions in this module are written for the
+-- parameterized label data form and thus take a Fmt argument for handling the
+-- parameterized label type.
+--
+-- When the parameterized label type is known to be 'BlockLabel', then ppLabel
+-- can be passed as this first argument to those pretty printing functions.
+--
+-- When the parameterized label type is not known (e.g. when implementing other
+-- libraries that utilize llvm-pretty), the library may need to invoke the pretty
```
RyanGlScott: Haddock nits:


```suggestion
-- parameterized label data form and thus take a 'Fmt' argument for handling the
-- parameterized label type.
--
-- When the parameterized label type is known to be 'BlockLabel', then 'ppLabel'
-- can be passed as this first argument to those pretty printing functions.
--
-- When the parameterized label type is not known (e.g. when implementing other
-- libraries that utilize @llvm-pretty@), the library may need to invoke the pretty
```

RyanGlScott (commented):
There are two parts to this commit:

1. Exporting `Fmt`.
2. Adding the `LabelPrinter` class.

(1) is entirely straightforward. I have some more questions about (2), however.

src/Text/LLVM/PP.hs
kquick: Ok...  Given that this whole module is about the `Pretty` implementation and the `Pretty` class is referenced in the example in these same Haddocks, I'm not sure I understand what you think is missing here; I've added a mention but let me know if your concern still isn't addressed.

I will say that this class is one of those where you won't need it until you do, and then you find it and it'll be "aha, that's exactly what I need".

kquick: Renamed to conform more closely with conventions here and in AST.hs.

src/Text/LLVM/PP.hs
kquick: Thanks.  Fixed.

src/Text/LLVM/PP.hs
RyanGlScott: Thanks. It's a bit of a shame that the name `ppLabel` is already taken, as that would have been perfect for this class method. (Alas, `ppLabel` is already taken for the function that pretty-prints `BlockLabel`s, and it's probably not worth a deprecation cycle at this point.)

RyanGlScott: In my mind, the code for implementing pretty-printing (i.e., all of the monomorphic `pp*` functions) is distinct from the type classes to provide ad hoc polymorphism for pretty-printing (i.e., the `Pretty` and `PrettyLabel` classes). As such, I find the sentence "Most of the pretty printing functions (based on `Pretty`) in this module [...]" to be a bit confusing, since the two types of code serve different purposes.

I think what I was looking for is a sentence or two after the code snippet (where you give an example of how to use `Pretty` and `PrettyLabel`) that says:

> Note that the `PrettyLabel` class is very similar to (but not quite the same thing as) the `Pretty` class. Unlike the `Pretty` class's `pretty` method, the `PrettyLabel` class's `ppLabel'` method is allowed to choose different pretty-printing approaches based on the implicit `Config` parameter that `Fmt` provides.

RyanGlScott (approved):

src/Text/LLVM/PP.hs
kquick: Typo from where I imported this from, thanks for catching that.

PR #196, opened 2026-04-22
Adds DISubrangeType (LLVM 21).

kquick:
This also updates pretty-printing to output negative values to be compatible with the LLVM tooling which expects to parse a signed 64-bit (instead of an unsigned 64-bit that will later be converted to signed).

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -1946,19 +1946,19 @@ data DIBasicType' lab = DIBasicType
 
 type DIBasicType = DIBasicType' BlockLabel
 
-data DISubrangeType' lab = DISubrangeType -- Added in LLVM 22
+data DISubrangeType' lab = DISubrangeType -- Added in LLVM 21
   { disrtName       :: Maybe String
   , disrtFile       :: Maybe (ValMd' lab)
   , disrtLine       :: Word32
   , disrtScope      :: Maybe (ValMd' lab)
   , disrtBaseType   :: Maybe (ValMd' lab) -- ^ a type
-  , disrtSize       :: Maybe (ValMd' lab) -- ^ in bits. signed constant, DIVariable, or DIExpression
+  , disrtSize       :: Maybe (ValMd' lab) -- ^ in bits. signed constant, DIVariable, DIGlobalVariable, or DIExpression
   , disrtAlign      :: Word64 -- ^ in bits
   , disrtFlags      :: DIFlags
-  , disrtLowerBound :: Maybe (ValMd' lab) -- ^ signed constant, DIVariable, or DIExpression
-  , disrtUpperBound :: Maybe (ValMd' lab) -- ^ signed constant, DIVariable, or DIExpression
-  , disrtStride     :: Maybe (ValMd' lab) -- ^ signed constant, DIVariable, or DIExpression
-  , disrtBias       :: Maybe (ValMd' lab) -- ^ signed constant, DIVariable, or DIExpression
+  , disrtLowerBound :: Maybe (ValMd' lab) -- ^ signed constant, DIVariable, DIGlobalVariable, or DIExpression
+  , disrtUpperBound :: Maybe (ValMd' lab) -- ^ signed constant, DIVariable, DIGlobalVariable, or DIExpression
+  , disrtStride     :: Maybe (ValMd' lab) -- ^ signed constant, DIVariable, DIGlobalVariable, or DIExpression
+  , disrtBias       :: Maybe (ValMd' lab) -- ^ signed constant, DIVariable, DIGlobalVariable, or DIExpression
   } deriving (Data, Eq, Functor, Generic, Ord, Show)
 
 type DISubrangeType = DISubrangeType' BlockLabel
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -179,6 +179,7 @@ module Text.LLVM.AST
   , DILocalVariable'(..), DILocalVariable
   , DISubprogram'(..), DISubprogram
   , DISubrange'(..), DISubrange
+  , DISubrangeType'(..), DISubrangeType
   , DISubroutineType'(..), DISubroutineType
   , DIArgList'(..), DIArgList
   , dwarf_DW_APPLE_ENUM_KIND_invalid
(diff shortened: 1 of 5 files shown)


src/Text/LLVM/AST.hs
```diff
@@ -1944,6 +1946,23 @@ data DIBasicType' lab = DIBasicType
 
 type DIBasicType = DIBasicType' BlockLabel
 
+data DISubrangeType' lab = DISubrangeType -- Added in LLVM 22
```
RyanGlScott: FWIW, https://github.com/llvm/llvm-project/commit/e298fc2da97120a30ee2f120ac184ab209fc1eb4 suggests that `DISubrangeType` was added in LLVM 21, not 22. (I didn't notice it when adding support for LLVM 21 since I was testing using C, not Ada, and the comments on that commit message suggest that this is primarily of use for Ada.)

src/Text/LLVM/AST.hs
```diff
@@ -1944,6 +1946,23 @@ data DIBasicType' lab = DIBasicType
...
+  { disrtName       :: Maybe String
+  , disrtFile       :: Maybe (ValMd' lab)
+  , disrtLine       :: Word32
+  , disrtScope      :: Maybe (ValMd' lab)
+  , disrtBaseType   :: Maybe (ValMd' lab) -- ^ a type
+  , disrtSize       :: Maybe (ValMd' lab) -- ^ in bits. signed constant, DIVariable, or DIExpression
```
RyanGlScott: I think this can also be a `DIGlobalVariable`, if I understand https://github.com/llvm/llvm-project/commit/3b90597c2ceaae86608214f6b62b43e55823102b correctly. (See also the comments on [this field](https://github.com/GaloisInc/llvm-pretty/blob/2982b117278d1563f068c2ebf0912a0e039a55df/src/Text/LLVM/AST.hs#L1934-L1937), which was also affect by the same LLVM commit.)

src/Text/LLVM/PP.hs
```diff
@@ -1503,6 +1506,26 @@ ppDIBasicType' pp bt = "!DIBasicType"
...
        )
 
+ppDISubrangeType' :: Fmt i -> Fmt (DISubrangeType' i)
+ppDISubrangeType' pp srt = "!DISubrangeType"
+  <> parens (mcommas
+       [     (("name:"     <+>) . doubleQuotes . text) <$> (disrtName srt)
+       ,     (("file:"     <+>) . ppValMd' pp) <$> (disrtFile srt)
+       , pure ("line:"     <+> integral (disrtLine srt))
+       ,     (("scope:"    <+>) . ppValMd' pp) <$> (disrtScope srt)
+       ,     (("size:"     <+>) . ppValMd' pp) <$> (disrtSize srt)
```
RyanGlScott, line 1516: For consistency, I think this should be printed using `ppSizeOrOffsetValMd'`. (See also [this](https://github.com/GaloisInc/llvm-pretty/blob/2982b117278d1563f068c2ebf0912a0e039a55df/src/Text/LLVM/PP.hs#L1489) similar code.)

src/Text/LLVM/PP.hs
```diff
@@ -1859,7 +1882,16 @@ ppInt64ValMd' canFallBack pp = go
...
+              ->
+                -- 64 bits is the largest Int, so no conversion needed.  However,
+                -- we intend to be compatible with LLVM's own tooling, and the
+                -- latter will parse these values as signed and will therefore
+                -- fail to parse a value with the high bit set.  Therefore this
+                -- must emit a signed value as well.
+                let v = fromInteger i :: Word64
+                in if testBit v 63
+                   then char '-' <> integer (toInteger $ complement v + 1)
+                   else integer i
```
RyanGlScott: I'm a bit uncomfortable with the fact that `llvm-pretty-bc-parser` is producing an `ValInteger` where the underlying `Integer` is in two's complement form, as this means that most downstream users of this metadata will need to perform the two's complement conversion (i.e., the `complement v + 1`) part themselves. Given that we know that we are parsing signed integers, I think it would be preferable to have `llvm-pretty-bc-parser` perform this conversion right after parsing it. That way, we can simply display the signed here with `integer`.

src/Text/LLVM/AST.hs
kquick: So it is.  I had based it on the fact that it followed the `llvmorg-21-init` tag, but hadn't seen that it was actually in llvm 21 tagged releases.  I verified that this is present in LLVM 21 but not LLVM 20 and changed the comments accordingly.

src/Text/LLVM/AST.hs
kquick: Updated.

src/Text/LLVM/PP.hs
kquick: I was hesitant to use that because that implied that this field could have a different representation pre-LLVM21, whereas this entire record does not exist pre-LLVM21.  So it's technically correct, but a little misleading.

I think it might be more accurate to change `ppSizeOrOffsetValMd'` to something like `ppOldSizeOrOffsetValMd'` or `ppSizePossiblyPreLLVM21'`. 

Your thoughts?

src/Text/LLVM/PP.hs
kquick: I had been avoiding that due to the much larger potential for impact on our other tools, but it is a more correct parse, so I'll make that change and we can address any issues as they arise.

src/Text/LLVM/PP.hs
RyanGlScott: Fair point. It appears that the behavior of `ppSizeOrOffsetValMd'` is equivalent to that of `ppValMd'` when LLVM >= 21 is in use, so I suppose we can just use the latter here.

I don't have too strong of an opinion on what to name `ppSizeOrOffsetValMd'`, although it's worth noting that the name is already quite long, and I'm a bit reluctant to make it even longer. I think the Haddocks accompanying it convey the subtleties pretty well.

src/Text/LLVM/PP.hs
kquick: Spot fixed.

src/Text/LLVM/PP.hs
```diff
@@ -1859,7 +1882,9 @@ ppInt64ValMd' canFallBack pp = go
           ValMdValue tv
             | PrimType (Integer _) <- typedType tv
             , ValInteger i <- typedValue tv
-              -> integer i  -- 64 bits is the largest Int, so no conversion needed
+              ->
+                -- 64 bits is the largest Int, so no conversion needed.
+                integer i
```
RyanGlScott: There's some needless whitespace churn here. Can you revert this back to the old code to make the diff simpler?

RyanGlScott (approved):
Make sure to mention these changes in the changelog.

src/Text/LLVM/PP.hs
kquick: Sure.  I had done this while I was in town to get the width under 80 characters, but that isn't a driving issue.  Reverted.

PR #198, opened 2026-04-25
Fix type pretty-printing for void pointers.

kquick:
It's possible (but not wise) to construct a `PtrTo PtrOpaque`, which will incorrectly be emitted with a following asterisk, resulting in something like `store ptr %0, ptr* %2, align 8`, which is invalid because ptr is void and does not allow additional pointer indirection.

This fix detects this construction and avoids the extra indirection, which will be rejected by the LLVM tools.

### DIFF
--- a/src/Text/LLVM/PP.hs
+++ b/src/Text/LLVM/PP.hs
@@ -477,7 +477,9 @@ ppType :: Fmt Type
 ppType (PrimType pt)     = ppPrimType pt
 ppType (Alias i)         = ppIdent i
 ppType (Array len ty)    = brackets (integral len <+> char 'x' <+> ppType ty)
-ppType (PtrTo ty)        = ppType ty <> char '*'
+ppType (PtrTo ty)        = ppType ty <> case ty of
+                                          PtrOpaque -> mempty
+                                          _ -> char '*'
 ppType PtrOpaque         = "ptr"
 ppType (Struct ts)       = structBraces (commas (map ppType ts))
 ppType (PackedStruct ts) = angles (structBraces (commas (map ppType ts)))


RyanGlScott:
I think I am on board with this approach, although I don't think this completely solves the problem of intermixing `PtrTo` and `PtrOpaque`.

The discussion in https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/262 is highly relevant to this change. Ideally, `llvm-pretty-bc-parser` wouldn't generate an `llvm-pretty` AST that contains `PtrTo PtrOpaque` in the first place, since that mixes pointer syntaxes from different LLVM versions that were never meant to be combined. That being said, accomplishing this is easier said than done. `llvm-pretty-bc-parser` aims to be backwards-compatible with multiple LLVM versions, and since there isn't a straightforward way to figure out whether a given LLVM bitcode file corresponds to a particular version, `llvm-pretty-bc-parser` sometimes has to guess what pointer syntax to use (see https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/262#issuecomment-1789408493), resulting in oddities like `PtrTo PtrOpaque`.

The change in this PR works around the problem by pretty-printing types like `ptr*` as `ptr`. Strictly speaking, this PR fixes https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/262, as `llvm-pretty-bc-parser` will no longer generate completely ill-formed types such as `ptr*`. That being said, this PR doesn't solve the more general problem of mixing the two pointer syntaxes in the same file. Even with this PR, there still situations where `llvm-pretty-bc-parser` will incorrectly generate ASTs involving both `PtrTo` and `PtrOpaque`. For instance, consider an LLVM bitcode file generated from this C program:

```c
int f(int x[2]) {
    return x[1];
}
```

If you compile this to bitcode using a recent version of Clang (using `clang -O2 -emit-llvm -c`) and disassemble it using `llvm-dis`, the output will look something like this:

```llvm
define i32 @f(ptr %x) {
  %p = getelementptr i8, ptr %x, i64 4
  %r = load i32, ptr %p, align 4
  ret i32 %r
}
```

On the other hand, if you round-trip the bitcode file through `llvm-pretty-bc-parser` and pretty-print it using `llvm-pretty`, the output will instead look like:

```llvm
define default i32 @f(ptr %x) {
; <label>: 0
  %p = getelementptr i8, ptr %x, i64 4
  %r = load i32, i8* %p, align 4
  ret i32 %r
}
```

Note that the `load` instruction now uses `i8*` instead of `ptr`! This is because the `load` instruction's bitcode stores the pointee type (`i8`) internally rather than the pointer type, so `llvm-pretty-bc-parser` has to guess whether to render the pointer type as `i8*` or `ptr`. For the sake of maximal backwards compatibility, `llvm-pretty-bc-parser` chooses `i8*`, but this has the strange side effect of using both `ptr` and `i8*` in the same file.

That being said, this problem is not as severe as the one identified in https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/262. While `llvm-as` will outright reject types such as `ptr*`, `llvm-as` will still accept programs such as the `@f` function above (it will convert `i8*` to `ptr` internally). As such, I think we can live with this issue for now, but longer-term, we should think about making `llvm-pretty-bc-parser` consistently generate one pointer syntax or the other.

src/Text/LLVM/PP.hs
```diff
@@ -477,7 +477,9 @@ ppType :: Fmt Type
 ppType (PrimType pt)     = ppPrimType pt
 ppType (Alias i)         = ppIdent i
 ppType (Array len ty)    = brackets (integral len <+> char 'x' <+> ppType ty)
-ppType (PtrTo ty)        = ppType ty <> char '*'
+ppType (PtrTo ty)        = ppType ty <> case ty of
+                                          PtrOpaque -> mempty
+                                          _ -> char '*'
```
RyanGlScott, line 482: This code definitely deserves a comment, as it's somewhat non-obvious why we'd need to collapse `ptr*` to `ptr`.

PR #199, opened 2026-05-26
Model `dead_on_return` parameter attribute (LLVM 21+)

AmeliaRose802:
Adds support for the `dead_on_return` parameter attribute introduced in LLVM 21. Without this, modern Clang/Rust LLVM bitcode that uses the attribute fails to round-trip through `llvm-pretty`.

1 commit, small AST addition + pretty-printing. Independent single-commit PR.

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -97,6 +97,7 @@ module Text.LLVM.AST
   , defFunType, addDefine
     -- * Function Attributes and attribute groups
   , FunAttr(..)
+  , ParamAttr(..)
     -- * Basic Block Labels
   , BlockLabel(..)
     -- * Basic Blocks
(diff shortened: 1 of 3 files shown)


langston-barrett (changes requested):
This actually appears to add many different constructions. `ParamAttr` itself does not appear in the `Module` AST. Can you clarify what this is supposed to be doing?

kquick:
See also the note on https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/366

AmeliaRose802:
Sorry about that! I didn't think draft PRs would be looked at till published. Was trying to decompose a big piece of work across submodules with AI and it all got away from me. Working on cleaning these up asap

AmeliaRose802:
Closing for now. Happy to revisit in a future toolchain-bump effort if/when SAW formally targets LLVM 21/22 bitcode. The branch will stay on my fork (`saw/dead-on-return`) for reference.

kquick:
The submissions are appreciated, and there's no way you could have known about the background work.  In general we prefer fairly focused submissions (one set of changes/features/fixes) per PR, with accompanying tests and references.  We're looking forward to ongoing collaboration, and would like to work with you to facilitate the ongoing work.

Can you perhaps share what your target is with this fleet of changes (beyond just completeness), and if there is a deadline you are working under that we should be aware of?

PR #200, opened 2026-05-26
Add Token primitive type support

AmeliaRose802:
Adds AST + pretty-printing for the LLVM `token` primitive type, used by SEH funclet bundles, coroutine intrinsics, and a few other LLVM features. Without this, bitcode that mentions `token` types fails to round-trip.

1 commit, AST-only addition. Filed as draft for upstream consideration — `saw/seh-ast-only` (the main SEH PR) does not require this since SEH funclet tokens can also be carried via the existing operand-bundle path. Open this PR mainly to surface the design choice for upstream feedback.

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -538,6 +538,7 @@ data PrimType
   | Integer Word32
   | FloatType FloatType
   | X86mmx
+  | Token
   | Metadata
     deriving (Data, Eq, Generic, Ord, Show, Lift)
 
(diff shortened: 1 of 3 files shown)


kquick:
viz https://llvm.org/docs/LangRef.html#token-type

This seems reasonable to add and I'm happy to approve and merge if marked non-draft.

AmeliaRose802:
Closing as redundant: PR #201 includes the same `Token` `PrimType` addition (with a Haddock comment) alongside the full SEH instruction set, so this standalone carve-out is no longer needed.

PR #203, opened 2026-07-06
Adds llvmModuleCombine functionality and deprecates the Module semigroup

kquick:
Note: this is transplanted to here from SAW as it properly belongs here (I plan to remove the SAW version in favor of this one).  Changes from the SAW version include using syb `everywhere` instead of lens `biplate` because this module doesn't use lens, deprecation of the Semigroup (for high-enough GHC versions), and tests.

### DIFF
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -247,7 +247,7 @@ data Module = Module
 instance
 #if __GLASGOW_HASKELL__ >= 910
   -- Deprecation of instances was added in GHC 9.10
- {-# DEPRECATED "Unsafe! use llvmModuleCombine instead" #-}
+ {-# DEPRECATED "Unsafe! Scheduled for removal: use llvmModuleCombine instead" #-}
 #endif
   Sem.Semigroup Module where
   m1 <> m2 = Module
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -247,7 +247,7 @@ data Module = Module
 instance
 #if __GLASGOW_HASKELL__ >= 910
   -- Deprecation of instances was added in GHC 9.10
- {-# DEPRECATED "Unsafe! use llvm-combine library instead" #-}
+ {-# DEPRECATED "Unsafe! use llvmModuleCombine instead" #-}
 #endif
   Sem.Semigroup Module where
   m1 <> m2 = Module
--- a/src/Text/LLVM/AST.hs
+++ b/src/Text/LLVM/AST.hs
@@ -10,6 +10,7 @@ incomplete: there are some values that new LLVM versions would accept but are
 not yet represented here.
 -}
 
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE ViewPatterns #-}
@@ -243,7 +244,12 @@ data Module = Module
   } deriving (Data, Eq, Ord, Generic, Show)
 
 -- | Combines fields pointwise.
-instance Sem.Semigroup Module where
+instance
+#if __GLASGOW_HASKELL__ >= 910
+  -- Deprecation of instances was added in GHC 9.10
+ {-# DEPRECATED "Unsafe! use llvm-combine library instead" #-}
+#endif
+  Sem.Semigroup Module where
   m1 <> m2 = Module
     { modSourceName = modSourceName m1 `mplus`   modSourceName m2
     , modTriple     = modTriple m1     <> modTriple     m2
@@ -259,9 +265,13 @@ instance Sem.Semigroup Module where
     , modComdat     = modComdat     m1 <> modComdat     m2
     }
 
-instance Monoid Module where
+instance
+#if __GLASGOW_HASKELL__ >= 910
+  -- Deprecation of instances was added in GHC 9.10
+  {-# DEPRECATED "Scheduled for removal: use emptyModule instead" #-}
+#endif
+  Monoid Module where
   mempty = emptyModule
-  mappend = (<>)
 
 emptyModule :: Module
 emptyModule  = Module
--- a/src/Text/LLVM/Combine.hs
+++ b/src/Text/LLVM/Combine.hs
@@ -1,0 +1,217 @@
+{-# LANGUAGE LambdaCase #-}
+
+{- |
+Module      : Text.LLVM.Combine
+Description : Combine LLVM Modules
+License     : BSD3
+Maintainer  : Kevin Quick <kquick@galois.com>
+Stability   : provisional
+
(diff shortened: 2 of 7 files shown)


src/Text/LLVM/AST.hs
```diff
@@ -243,7 +244,12 @@ data Module = Module
   } deriving (Data, Eq, Ord, Generic, Show)
 
 -- | Combines fields pointwise.
-instance Sem.Semigroup Module where
+instance
+#if __GLASGOW_HASKELL__ >= 910
+  -- Deprecation of instances was added in GHC 9.10
+ {-# DEPRECATED "Unsafe! use llvm-combine library instead" #-}
```
RyanGlScott: What is the `llvm-combine` library? Is this meant to reference `llvmModuleCombine` instead?

Also, is this scheduled for removal like the `Monoid` instance is? If so, worth noting here in the deprecation text.

src/Text/LLVM/Combine.hs
```diff
@@ -0,0 +1,217 @@
+{-# LANGUAGE LambdaCase #-}
+
+{- |
+Module      : Text.LLVM.Combine
+Description : Combine LLVM Modules
+License     : BSD3
+Maintainer  : Kevin Quick <kquick@galois.com>
+Stability   : provisional
+
+Thos module provides the ability to smash together LLVM Module specifications to
```
RyanGlScott:
```suggestion
This module provides the ability to smash together LLVM Module specifications to
```

src/Text/LLVM/Combine.hs
```diff
@@ -0,0 +1,217 @@
...
+provide the ability to load separate LLVM Modules (e.g. bitcode files) and
```
RyanGlScott:
```suggestion
Thos module provides the ability to smash together LLVM 'Module' specifications to
provide the ability to load separate LLVM 'Module's (e.g. bitcode files) and
```

src/Text/LLVM/Combine.hs
```diff
@@ -0,0 +1,217 @@
...
+--
+-- 2. The `llvm-link` tool will occasionally rewrite calls to llvm intrinsics to
+--    explicitly add the default personality specification.  For example,
+--    `llvm.stacksave` may be rewritten to `llvm.stacksave.p0`.  Because these
+--    are intrinsics, this should not have any significant impact on the result,
+--    but `llvmModuleCombine` does not perform this naming update.
+--
+-- 3. External declaration resolution is type independent and only name
+--    sensitive.  If Module A has an external declaration `declare @f(i32 x)` and
+--    Module B has a definition `define @f(float x)`, this `llvmModuleCombine`
```
RyanGlScott:
```suggestion
-- | Combines LLVM 'Module's into a single, composite 'Module'.  This is akin to
-- linking, but just from the perspective of what is needed for program analysis.
--
-- This differs from `llvm-link` in the following known ways:
--
-- 1. The `llvm-link` tool uses structural typing resolution: if two modules
--    each have a type with the same structure, the resulting module will only
--    have one type; the name from one of the modules is chosen and all
--    references to the typename in the other module will be rewritten to the
--    first module.
--
--    The `llvmModuleCombine` takes a slightly different approach: types are not
--    structurally coalesced, but this means that type names are deconflicted by
--    adding a numbered suffix.  This still requires modifying the type name
--    throughout that module, but (a) there are probably fewer type name
--    conflicts than structural equivalences, and (b) the original name is still
--    part of the new name which maintains origin information.
--
-- 2. The `llvm-link` tool will occasionally rewrite calls to llvm intrinsics to
--    explicitly add the default personality specification.  For example,
--    `llvm.stacksave` may be rewritten to `llvm.stacksave.p0`.  Because these
--    are intrinsics, this should not have any significant impact on the result,
--    but `llvmModuleCombine` does not perform this naming update.
--
-- 3. External declaration resolution is type independent and only name
--    sensitive.  If 'Module' A has an external declaration `declare @f(i32 x)` and
--    'Module' B has a definition `define @f(float x)`, this `llvmModuleCombine`
```

src/Text/LLVM/Combine.hs
```diff
@@ -0,0 +1,217 @@
...
+--
+-- This differs from `llvm-link` in the following known ways:
+--
+-- 1. The `llvm-link` tool uses structural typing resolution: if two modules
+--    each have a type with the same structure, the resulting module will only
+--    have one type; the name from one of the modules is chosen and all
+--    references to the typename in the other module will be rewritten to the
+--    first module.
+--
+--    The `llvmModuleCombine` takes a slightly different approach: types are not
```
RyanGlScott:
```suggestion
--    `llvmModuleCombine` takes a slightly different approach: types are not
```

src/Text/LLVM/Combine.hs
```diff
@@ -0,0 +1,217 @@
...
+                then error $ "Unable to generate unique type name for " <> b
+                else renameType m t b $ succ n
+           else everywhere (mkT (chngType (typeName t) newName)) m
+      chngType oldName newName n = bool n newName $ n == oldName
+  in foldl resolveTypeConflict inpMod (inpMod ^. modTypesLens)
+
+
+-- | A Define takes precedence over a Declare.  When combining modules, module A
+-- may Declare a function that is handled by a Define in module B, so get rid of
+-- the Declare when putting A and B together.
```
RyanGlScott:
```suggestion
-- | A 'Define' takes precedence over a 'Declare'.  When combining modules, module A
-- may 'Declare' a function that is handled by a 'Define' in module B, so get rid of
-- the 'Declare' when putting A and B together.
```

src/Text/LLVM/Combine.hs
```diff
@@ -0,0 +1,217 @@
...
+
+removeDefined :: Define -> [Declare] -> [Declare]
+removeDefined def = filter ((def ^. defNameLens /=) . view decNameLens)
+
+
+-- | Module A and Module B may have a Define with the same name (Symbol).  This
+-- is normal when linking multiple modules together, and is resolved by linkers
+-- as guided by the Linkage information for the two Definitions, usually by
+-- either renaming or merging.
```
RyanGlScott:
```suggestion
-- | 'Module' A and 'Module' B may have a 'Define' with the same name ('Symbol').  This
-- is normal when linking multiple modules together, and is resolved by linkers
-- as guided by the 'Linkage' information for the two 'Definition's, usually by
-- either renaming or merging.
```

src/Text/LLVM/AST.hs
kquick: Ahh yes, it was initially in a separate library, but that was too abstruse, and since `syb` has `everywhere` (instead of `lens` `biplate`) I could move it here.

kquick:
I addressed all the Haddock and deprecation suggestions, and also pushed a change removed the internal uses of the deprecated Semigroup instance.  It's again ready for review.

RyanGlScott (approved):

